PR #49, opened 2021-03-11
basic resume support

eaceaser:
This adds basic support for parsing the resume kernel parameter and setting `/sys/power/resume` appropriately. It works for me, but I haven't tested all combinations of argument formats yet (e.g. a LABEL device selector) -- I wanted to get this PR in front of people early because it kinda feels like a general refactor of `devAdd` might be in order if this approach I'm taking is what we want.

Couple notes;
* I chose to add the resume handling to the `devAdd` handler, because that's also where luks handling that similarly needs to lookup devices lives. I'm not sure if this is the best approach -- seems like we could maybe do the resume handling after we've mounted the root volume, and just no-op if the device hasn't been added by that point?
* If doing this logic in the `devAdd` handler makes sense, it's currently a bit unwieldy because there are a few return points from the function, and currently it seems like we'd have to call the `resume` handler at each point. I'd love to discuss ways to refactor the function if necessary.

### DIFF
--- a/init/main.go
+++ b/init/main.go
@@ -113,6 +113,9 @@ func devAdd(syspath, devname string) error {
 
 	devpath := path.Join("/dev", devname)
 	if devpath == cmdroot {
+		if err := resume(devpath); err != nil {
+			return err
+		}
 		return mountRootFs(devpath)
 	}
 
@@ -137,6 +140,10 @@ func devAdd(syspath, devname string) error {
 			return err
 		}
 
+		if err := resume(devpath); err != nil {
+			return err
+		}
+
 		if dmPath == cmdroot {
 			return mountRootFs(dmPath)
 		}
@@ -147,10 +154,11 @@ func devAdd(syspath, devname string) error {
 		return fmt.Errorf("%s: %v", devpath, err)
 	}
 
-	if strings.HasPrefix(cmdroot, "UUID=") && info.uuid == strings.TrimPrefix(cmdroot, "UUID=") {
-		return mountRootFs(devpath)
+	if err := resume(devpath); err != nil {
+		return err
 	}
-	if strings.HasPrefix(cmdroot, "LABEL=") && info.label == strings.TrimPrefix(cmdroot, "LABEL=") {
+
+	if blkInfoMatchesArg(info, cmdroot) {
 		return mountRootFs(devpath)
 	}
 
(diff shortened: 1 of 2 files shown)


init/main.go
```diff
@@ -697,6 +705,57 @@ func checkIfInitrd() error {
...
+		}
+
+		if blkInfoMatchesArg(info, devpath) {
+			resumeDevice = devpath
+			goto resume
+		}
+	}
+
+	return nil
+resume:
```
eaceaser, line 744: i dont love this but if there's a larger refactor in the cards I'll address whether or not a goto makes sense here later

init/main.go
```diff
@@ -697,6 +705,57 @@ func checkIfInitrd() error {
...
+		return true
+	}
+	if strings.HasPrefix(arg, "LABEL=") && info.label == strings.TrimPrefix(arg, "LABEL=") {
+		return true
+	}
+
+	return false
+}
+
+func resume(devpath string) error {
```
anatol, line 719: does this patch handles suspend-on-disk only? is there anything special need to be to handle suspend-to-ram functionality?

init/main.go
```diff
@@ -697,6 +705,57 @@ func checkIfInitrd() error {
...
+	return nil
+resume:
+	major, minor, err := deviceNo(resumeDevice)
+	if err != nil {
+		return err
+	}
+
+	rd := fmt.Sprintf("%d:%d", major, minor)
+	debug("resume matched: %s=%s, writing %s", resume, resumeDevice, rd)
+	if err := os.WriteFile("/sys/power/resume", []byte(rd), 0644); err != nil {
```
anatol, line 752: where I can find more information about this "echo $diskno > /sys/power/resume" operation. What exactly kernel does here? How does suspend-to-XXX implemented in Linux?

init/main.go
anatol, line 719: Is there anything need to be done by booster at suspend time? Or it is assumed that the host software (e.g. systemd) handles it without involving booster.

init/main.go
```diff
@@ -697,6 +705,57 @@ func checkIfInitrd() error {
...
+	}
+	if strings.HasPrefix(arg, "LABEL=") && info.label == strings.TrimPrefix(arg, "LABEL=") {
+		return true
+	}
+
+	return false
+}
+
+func resume(devpath string) error {
+	resume, ok := cmdline["resume"]
```
anatol, line 720: Please update docs/manpage as well

anatol (commented):
Thank you for your change @eaceaser

I never used the `resume` functionality. I do not know anything about it but I eager to learn more. A couple of questions below.

init/main.go
eaceaser: This is purely for suspend-to-disk. Suspend-to-ram doesn't go through a reboot, so it's handled purely by the running kernel.

init/main.go
eaceaser: A couple resources:

* https://wiki.archlinux.org/index.php/Power_management/Suspend_and_hibernate#Required_kernel_parameters has a summary of what's going on here

* From the [kernel docs](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html):
   >>> 
        resume=         [SWSUSP]
                        Specify the partition device for software suspend
                        Format:
                        {/dev/<dev> | PARTUUID=<uuid> | <int>:<int> | <hex>}

        resume_offset=  [SWSUSP]
                        Specify the offset from the beginning of the partition
                        given by "resume=" at which the swap header is located,
                        in <PAGE_SIZE> units (needed only for swap files).
                        See  Documentation/power/swsusp-and-swap-files.rst

* for reference, [mkinitcpio](https://github.com/archlinux/mkinitcpio/blob/master/hooks/resume#L25-L32) has similar logic, as does [dracut](https://github.com/dracutdevs/dracut/blob/master/modules.d/95resume/resume.sh#L17)

One thing that I noticed when implementing this is that it appears that don't need to handle the `resume.offset` kernel parameter in the initrd, just the `resume` parameter - the kernel seems to parse that from its commandline itself when resuming, but it needs `/sys/power/resume` to have the major/minor of the block device set already for the resume to occur.

init/main.go
eaceaser: nope, as long as `/sys/power/resume` and `/sys/power/resume_offset` are set correctly, when suspend-to-disk happens the kernel handles everything

init/main.go
eaceaser: added some docs

anatol:
Thank you for your work.

I just finished the `init` code refactoring that among other things should address duplication in the 'match block device' code.

I rebased your change on top of it and pushed both your changes and my refactoring to `resume` branch. Could you please check if it works as expected.

anatol:
Also it would be really great to have an integration test for resume functionality. Is there a way to create a disk with "suspended" state and then boot using it?

anatol:
D'oh messed up with git commit author. I am going to fix it really quick.

anatol:
Hi Ed, I would like to move with this change forward. Had you chance to review/verify the rebased commit?

eaceaser:
@anatol Hey, sorry, I've been dealing with some life stuff over the past week or so. I haven't forgotten bout this, I'll find some time to rebase and test.

eaceaser:
@anatol Just tested https://github.com/anatol/booster/commit/8175bbeec05aa47c9ae7bef060aac86c100c4525 -- it appears to work perfectly. Sorry again that i haven't had a ton of time these past few weeks to shepherd this so I appreciate your efforts.

anatol:
No worries, you did a great work. Thank you very much for your contribution.

PR #50, opened 2021-03-11
Parse quoted UUID from cmdline

williambanfield:
This pull requests attempts to partially remedy #6 

This change adds a `parseUUIDFromCmdRoot` function that parses the cmdroot UUID field and a set of tests for the logic
The change adds a regex for the UUID that allows a set of quotes around the UUID field to optionally be added. The regular expression expects a UUID in the [canonical format](https://en.wikipedia.org/wiki/Universally_unique_identifier#Format), 8-4-4-4-12. 

Since I don't have `root=UUID=""` filesystem using lux, I haven't been able to end-to-end test this on my machine, but happy to try to get a vm setup to attempt this if we think that's best.

Happy to make any changes wanted here if it seems this could be helpful.

### DIFF
--- a/init/main.go
+++ b/init/main.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


billybanfield:
Hey @anatol, do you think this would be helpful? I'd be happy to update in different ways if you think it would be.

anatol:
Thanks for your PR and sorry for the delay. The general idea looks good to me. Though I prefer to have 2 different, atomic functions:
1. validate UUID format
2. remove enclosing quotes

The funcion 2) can be later used to remove quotes over e.g. `LABEL=` parameter.

A few notes wrt the implementation:

1. UUID should be able to accept both upper and lower case hexadecimals
2. quotes should be remove only if they pared i.e. if a quote only at at the beginning of the UUID then it is an invalid parameter and should be rejected.

In fact I have a refactoring work sitting in my local branch that tries to cleanup the UUID handling and I already have this UUID format validation check. It covers both issue #6 and #8 functionality.

Issue #57, opened 2021-03-18
strip more section from ko and binary.


oech3:
ko and binary is stripped if `strip: true`.
But there are more strippable sections:
.note.* and .comment from ko.
.note.* and .go.buildinfo and .gosymtab from init.

In my test,uncompressed img decreased 8KB+ from  `strip: true` uncompressed img.

anatol:
Currently booster uses `strip --strip-unneeded` for dynamic libs (including *.ko modules) and `strip --strip-all` for standalone ELF files. What are the `strip` flags that allow to remove more data from the output elf files?

I was also thinking about implementing strip functionality purely in go. This will allow to have a stricter control over what we do with ELF files. Plus it will be faster and avoids having "Too many open files" error due to many `strip` invocations that happen simultaneously.

Is there any good documentation that explains what exactly `strip` does? Does it only remove specific sections based on its name? Or it performs additional operations with symbols?

oech3:
`strip --strip-unneeded -R .note.* -R .comment *.ko`
`strip -R .note.* -R .go.buildinfo -R .gosymtab init`
-R is same for --remove-section=

anatol:
Thanks. Added.

oech3:
Sorry for commenting after this issue closed.
It seems __bug_table and .rela__bug_table is also strippable from ko.
I don't know every section unneeded for ko.

I could not find documentation what strip do.
But https://github.com/NonerKao/go-binutils is binutils(including strip) implementation in Go which is not actively developed.

oech3:
.BTF and *orc_unwind* seems used for debug.sucseed to boot if they was stripped.

anatol:
> But https://github.com/NonerKao/go-binutils is binutils(including strip) implementation in Go which is not actively developed.

Thanks will look at it.

LLVM strip implementation could be also used as a source of information about strip functionality. https://github.com/llvm/llvm-project/blob/main/llvm/tools/llvm-objcopy/ELF/ELFObjcopy.cpp

> It seems __bug_table and .rela__bug_table is also strippable from ko.

It looks like this section is related to https://github.com/torvalds/linux/blob/master/lib/bug.c this code enabled bug IP address to filename:lineno matching for debug purposes. This feature is optional and I think it can be dropped from the *.ko.

oech3:
How about 
> .BTF and *orc_unwind* seems used for debug.sucseed to boot if they was stripped.
Sorry if you didn't miss this comment or just investigating.

[...]

PR #65, opened 2021-03-26
Strip ORC unwinder related section from ELF

oech3:
Strip .orc_unwind_ip .rela.orc_unwind_ip .orc_unwind from ELF binaries
https://www.kernel.org/doc/html/latest/x86/orc-unwinder.html
#57


anatol:
Thanks for the contribution!

PR #98, opened 2021-09-06
Full Discoverable Partitions Specification compatibility (WIP)

rdvdev2:
The current booster [Discoverable Partitions Specification](https://systemd.io/DISCOVERABLE_PARTITIONS/) support is pretty basic and doesn't work for every case. This PR aims to fully support the specification, allowing to auto-discover encrypted partitions or mount root read-only, among other features.

_WARNING: This PR is really a WIP, and not properly tested.  At the moment, this PR is provided to get feedback while developing the feature(s)._

Progress:
- [x] Unencrypted root support
- [x] Encrypted root support (LUKS)
- [x] Scan only the disk containing the used ESP
- [ ] Honour the GPT flags (read-only mounting and partition skipping)
- [ ] Dm-crypt? (Not familiar with it neither sure if Arch or booster support it)
- [ ] Documentation?
- [ ] Tests? (Possibly on the GPT flag reading process)
- [ ] Works on VM (Not tested yet)
- [ ] Works on real hardware (Not tested yet)


init/deviceref.go
```diff
@@ -3,6 +3,8 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"github.com/foxboron/go-uefi/efi/attributes"
```
anatol, line 6: third-party imports should be at the bottom of this section divided by an empty line

init/deviceref.go
```diff
@@ -86,24 +88,87 @@ func (d *deviceRef) resolveFromGptTable(devPath string, t []gptPart) *deviceRef
...
+
+	debug("Scanning %s for the root partition. Used ESP was found in this device.", devPath)
+
+	for _, p := range t {
+		if bytes.Equal(rootTypeGuid, p.typeGuid) {
+			// TODO: Honour GPT flags (read only & no auto)
+			rootDevName := calculateDevName(devPath, p.num)
+			debug("Found root partition: %s.", rootDevName)
+			info, err := readBlkInfo(rootDevName)
+			if err == nil {
```
anatol, line 117: this condition should be flipped to make the error handling more go-idiomatic:

if err != nil {
  debug("foo")
  continue
}

init/deviceref.go
```diff
@@ -86,24 +88,87 @@ func (d *deviceRef) resolveFromGptTable(devPath string, t []gptPart) *deviceRef
...
-			gptType, err := parseUUID(autodiscoveryGuid)
+		if enableAutodetect && efiVarsAvailable {
+			debug("%s= param is not specified. Use GPT partition autodiscovery.", name)
+
+			_, data, err := attributes.ReadEfivarsWithGuid("LoaderDevicePartUUID", *util.StringToGUID("4a67b082-0a4c-41cf-b6c7-440b29bb8c4f"))
+			if err != nil {
+				return nil, err
+			}
+
+			espUuid, err := util.ParseUtf16Var(data)
```
anatol, line 166: We already have a function to convert utf16 into a string - `fromUnicode16`. Let's use own implementation instead of this third party util function.

anatol (commented):
The idea of implementing discoverable partitions great. The code looks clean, good work! I left a few comments about this implementation.

It needs more tests and I can help you with it if needed (setting up unit tests is currently a bit tricky).

init/main.go
```diff
@@ -619,6 +622,15 @@ func boost() error {
 		return err
 	}
 
+	// Mount efivarfs if running in EFI mode
+	if _, err := os.Stat("/sys/firmware/efi"); errors.Is(err, os.ErrNotExist) {
```
anatol, line 626: BTW what is the difference of this construction from `os.IsNotExist(err)`?

rdvdev2: It is the same, wrote it this way as [go documentation](https://pkg.go.dev/os@go1.17#IsNotExist) recommended it. If it is preferred I can change it to `os.isNotExist(err)`

anatol: Ack. Let's follow docs recommendation. It says `errors.Is(err, fs.ErrNotExist)` should be used.

TODO for myself: change other places in booster and use `errors.Is` to be consistent with the documentation.

init/luks.go
```diff
@@ -145,35 +145,39 @@ func luksOpen(dev string, name string) error {
...
+	return "", nil
+}
+
+func handleLuksBlockDevice(info *blkInfo, devpath string) error {
+	name, err := parseLuksParameters(info)
+	if err != nil {
+		return err
+	}
+
+	if name != "" {
```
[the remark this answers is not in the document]
rdvdev2: It can happen if the kernel parameters don't refer to the tested device (for example encrypted /home) or if no luks parameters were set at all.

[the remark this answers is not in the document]
anatol: ack, let's flip it and handle error case inside the `if{}` branch.

```go
if name == "" {
  return nil
}

// .. the rest of the code
```

rdvdev2:
> It needs more tests and I can help you with it if needed (setting up integration tests is tricky currently).

I don't really know how testing works with this project, so help would be really appreciated.

init/main.go
```diff
@@ -619,6 +622,15 @@ func boost() error {
...
+		if err := mount("efivarfs", "/sys/firmware/efi/efivars", "efivarfs", unix.MS_NOSUID|unix.MS_NOEXEC|unix.MS_NODEV, ""); err != nil {
```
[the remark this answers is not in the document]
anatol: See`moveMountpointsToHost()` function. First it tries to move the mountpoint to host. If it does not work (e.g. new root does not contain the directory) then it unmounts the filesystem from initram fs tree.

The `moveMountpointsToHost()` should handle `/sys/firmware/efi` mountpoint as well. I wonder if `MS_MOVE` and `MNT_DETACH` handle mountpoints recursively.

anatol:
> I don't really know how testing works with this project, so help would be really appreciated.

The integration test are located at `tests` directory. Each test contains from 2 parts:
 1. a bash script that creates a host image (see `tests/generate_asset_*.sh`)
 2. golang test that runs QEMU with this host image

If you can take care of 1 then I'll add QEMU tests for it. See `generate_asset_gpt.sh` that creates an image with a GPT and multiple partitions. You can even run this script from you shell with `OUTPUT=assets/gpt.img FS_UUID=e5404205-ac6a-4e94-bb3b-14433d0af7d1 FS_LABEL=newpart bash -x -o errexit ./generate_asset_gpt.sh` it will generate a valid image at `assets/gpt.img`. You probably want something similar to initialize an GPT with particular partition UUIDs.

As of formatting a partition with LUKS check `generate_asset_luks.sh`.

init/main.go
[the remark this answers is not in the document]
rdvdev2: I've been reading through the documentation and I think that moving the mountpoint isn't needed at all. It looks like systemd handles all the mountpoints but `/run`.  Take a look at the [specification](https://systemd.io/INITRD_INTERFACE/), it may even be the case that the current code on `moveMountpointsToHost()` is mostly unneeded.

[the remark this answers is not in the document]
anatol: It actually used to be like mentioned in this specification. But it turns out that other init systems like `runit` expect `/dev/` to be mounted at the boot time. See https://github.com/anatol/booster/issues/92 and https://github.com/anatol/booster/commit/4c9505873ec5ecde77ed749091ce7bd106b5b401

I am fine to leave this code as-is and wait for integration tests results to see if runit tests pass.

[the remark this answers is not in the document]
rdvdev2: Then I'll try to get unit tests ready today and see if it works.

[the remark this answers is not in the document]
anatol: I ran your PR with VoidLinux and ArchLinux+btrfs tests and they passed.

anatol:
Hey I just want to check with you how it is going. Do you need any help with testing?

rdvdev2:
I'm sorry for the inactivity, last week I started university and I've been really busy since then. Now that I have some free days I'll hopefully be able to continue working on this. I'll let you know if I have any problems with the testing setup.

init/main.go
anatol, line 626: I believe it should be `!errors.Is(err, os.ErrNotExist)`. `if` body should be executed if the efi dir not-not exists (i.e. exists).

anatol:
I pulled the latest changes from this PR and rebased on top of `wip` branch.

anatol:
I tried to ran a test that uses autodiscovery (e.g. `go test -v -run TestBooster/Gpt.Autodiscovery`) and I see that it requires EFI mode for QEMU now. One question here (and sorry if it is dumb one) - does autodiscovery work *without* EFI?

I enable QEMU EFI mode with  `-bios /usr/share/ovmf/x64/OVMF.fd` parameters and now my test fails with `open /sys/firmware/efi/efivars/LoaderDevicePartUUID-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f: no such file or directory`. Is this efi var required for autodiscovery? What if it does not exist at a machine and no `root=` specified, should booster fail into emergency shell? Or is there a way to recover from this situation?

rdvdev2:
The EFI var is needed to know which physical disk contains the auto-detected system. I think this is done to avoid loading the kernel from a LiveUSB and the root of a locally installed system (and other probably breaking situations). Knowing this, I'm not sure if it would be a good idea to go ahead and autodetect on all drives, unless there's another way to detect where the kernel was loaded from.

Maybe we could still autodetect if there's only one disk that follows the specification or only one of them contains partitions for the system architecture, but in any other case, I think that falling back to the emergency shell would be the most sensible option. Let me know what you think.

rdvdev2:
The EFI var is needed to know which physical disk contains the auto-detected system. I think this is done to avoid loading the kernel from a LiveUSB and the root of a locally installed system (and other probably breaking situations). Knowing this, I'm not sure if it would be a good idea to go ahead and autodetect on all drives, unless there's another way to detect where the kernel was loaded from.

Maybe we could still autodetect if there's only one disk that follows the specification or only one of them contains partitions for the system architecture, but in any other case, I think that falling back to the emergency shell would be the most sensible option. Let me know what you think.

init/deviceref.go
```diff
@@ -45,27 +49,27 @@ func (d *deviceRef) matchesBlkInfo(info *blkInfo) bool {
...
-	}
-
 	for _, p := range t {
 		switch d.format {
-		case refGptType:
-			if bytes.Equal(d.data.(UUID), p.typeGUID) {
-				return &deviceRef{refName, calculateDevName(devPath, p.num)}
+		case refEspUUID:
+			if bytes.Equal(d.data.(UUID), p.uuid) {
+				return d.autodetectRoot(devPath, t)
```
anatol, line 72: The `autodetectRoot() function does not accept a partition but instead gets whole table. For me it looks like the `autodetectRoot()` should be split into 2 parts

1. calculate `rootTypeGuid` - move outside of the loop
2. `if bytes.Equal(rootTypeGuid, p.typeGUID) {` - should be called inside the loop

init/deviceref.go
```diff
@@ -86,24 +90,87 @@ func (d *deviceRef) resolveFromGptTable(devPath string, t []gptPart) *deviceRef
...
 func parseDeviceRef(name, param string, enableAutodetect bool) (*deviceRef, error) {
 	if param == "" {
 		// try to auto-discover gpt partition https://www.freedesktop.org/wiki/Specifications/DiscoverablePartitionsSpec/
-		if autodiscoveryGUID, ok := autodiscoveryGptTypes[runtime.GOARCH]; enableAutodetect && ok {
-			debug("%s= param is not specified. Use GPT partition autodiscovery with guid type %s", name, autodiscoveryGUID)
-			gptType, err := parseUUID(autodiscoveryGUID)
+		if enableAutodetect && efiVarsAvailable {
+			debug("%s= param is not specified. Use GPT partition autodiscovery.", name)
+
+			_, data, err := attributes.ReadEfivarsWithGuid("LoaderDevicePartUUID", *util.StringToGUID("4a67b082-0a4c-41cf-b6c7-440b29bb8c4f"))
```
anatol, line 163: where does `LoaderDevicePartUUID` requirement come from? I checked https://www.freedesktop.org/wiki/Specifications/DiscoverablePartitionsSpec/ and do not see any mention of this efi var.

anatol: Ok, found it here `the first partition with this type UUID on the disk containing the active EFI ESP is automatically mounted to the root directory /`

anatol: So basically we need to find a **disk** that contains the active EFI ESP. Is reading `LoaderDevicePartUUID` efivar and then iterating all GPT partitions the only way to do it?

anatol:
@rdvdev2 how do you feel if i pick this PR and complete it? I would like to do some refactoring of booster code along the way. I'll keep you as an author of this change.

rdvdev2:
> @rdvdev2 how do you feel if i pick this PR and complete it? I would like to do some refactoring of booster code along the way. I'll keep you as an author of this change.

Go ahead, I think I don't have enough spare time to work on this anymore. Thanks for understanding!

anatol:
@rdvdev2 I reworked your patches, made extra changes to booster codebase and added support for GPT partition attributes. I pushed the changes to `wip` branch. Please test it when you have a chance.

rdvdev2:
Thank you too for your help (and sorry for not being able to finish this myself)

PR #107, opened 2021-09-28
Don't load network modules unless network is enabled

codicodi:
In the common (and default) case when network is disabled, there is no need for any ethernet drivers. This can save a lot of firmware in some cases.

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -19,6 +19,7 @@ type generatorConfig struct {
 	networkStaticConfig     *networkStaticConfig
 	networkActiveInterfaces []net.HardwareAddr
 	universal               bool
+	defaultModules          []string
 	modules                 []string // extra modules to add
 	modulesForceLoad        []string // extra modules to load at the boot time
 	compression             string
@@ -73,7 +74,6 @@ var defaultModulesList = []string{
 	"kernel/crypto/",
 	"kernel/drivers/input/serio/",
 	"kernel/drivers/input/keyboard/",
-	"kernel/drivers/net/ethernet/",
 	"kernel/drivers/md/",
 	"kernel/drivers/char/tpm/",
 	"kernel/drivers/usb/host/",
@@ -105,6 +105,11 @@ func generateInitRamfs(conf *generatorConfig) error {
 
 	if err := img.appendExtraFiles(conf.extraFiles); err != nil {
 		return err
+	}
+
+	conf.defaultModules = defaultModulesList
+	if conf.networkConfigType != netOff {
+		conf.defaultModules = append(conf.defaultModules, "kernel/drivers/net/ethernet/")
 	}
 
 	if conf.enableLVM {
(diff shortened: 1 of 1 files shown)


generator/generator.go
```diff
@@ -107,6 +106,10 @@ func generateInitRamfs(conf *generatorConfig) error {
 		return err
 	}
 
+	if conf.networkConfigType != netOff {
+		defaultModulesList = append(defaultModulesList, "kernel/drivers/net/ethernet/")
```
anatol, line 110: `defaultModulesList` is a global variable and not suppose to be modified by a function. Activate the drivers the same way as mdraid does later:

```go
kmod.activateModules(true, true, "kernel/drivers/net/ethernet/")
```

codicodi: I've tried using `kmod.activateModules()` even before submitting this PR, but module dependencies weren't resolving correctly.
I went with new property for config, seems to work fine.

anatol:
Thank you for the patch. Now it looks good.

So I ran test and I see that with test `TestBooster/Vfio` starts failing for me. Booster actually mounts root successfully and gets into userspace, but any attempt to ssh to the virtual machine times out. It might be because the network state has changed somehow, though I would expect that userspace udev handles network initialization the same what booster did previously.

Anyway here is the test output in case if you see anything abnormal here https://gist.github.com/anatol/39b12483f7fae002690310a5532f9381

The image is configured here https://github.com/anatol/booster/blob/master/tests/generate_asset_archlinux_ext4.sh

anatol:
For comparison here is output of a passing test without this patch https://gist.github.com/anatol/d11e20a65ee48914f23fd90dac6a14a3

The userspace has following text
```
         Starting Network Configuration...
[    1.938800] e1000 0000:00:03.0 ens3: renamed from eth0
[  OK  ] Found device /dev/ttyS0.
[  OK  ] Started Network Configuration.
[    1.978195] e1000: ens3 NIC Link is Up 1000 Mbps Full Duplex, Flow Control: RX
[    1.978553] IPv6: ADDRCONF(NETDEV_CHANGE): ens3: link becomes ready
[  OK  ] Reached target Network.
```

codicodi:
I think I know what's going on. Generated images don't have any kernel installed and rely on drivers provided by booster.

codicodi:
I guess it's not viable to simply install kernel packages inside images since the images should work with any kernel.
I'm not very familiar with qemu, but is there a way to share `/usr/lib/modules` beetwen host and vm?

anatol:
> I think I know what's going on. Generated images don't have any kernel installed and rely on drivers provided by booster.

Spot on! Yeah images do not have linux-modules as tests use hosts' kernel that gets changing over time.

> I'm not very familiar with qemu, but is there a way to share /usr/lib/modules between host and vm?

I think getting host modules directory mounted to qemu would be ideal solution. I am not sure how to do it though.

> the images should work with any kernel.

Yes, one nice thing about the integration tests that it iterates over all kernels I have installed and checks that booster works with it.

For now I'll just add `e1000` module to tests that require SSH. But this mounting question needs to be researched.

anatol:
> I've tried using kmod.activateModules() even before submitting this PR, but module dependencies weren't resolving correctly.

Oh, it is a bug at booster. `activateModules()` activates (adds modules to the image) and finalizes (computes all dependencies of the modules) in one step. These operations (activation and finalization) need to be separated. I will look at it once I merge your PR.

anatol:
I pulled your change into `wip` branch. Please check if everything looks good.

codicodi:
Yep, I built the wip branch and it works fine, while not pulling >20 Realtek firmware files into the initramfs. Yay!

> For now I'll just add e1000 module to tests that require SSH.

Ah, I was about to propose the same workaround but I wasn't sure which module was needed.

anatol:
This PR has been merged to master as https://github.com/anatol/booster/commit/91f1da53071a1791ad8b7124e2e68574b9086fa4

PR #108, opened 2021-09-30
Add Void Linux instructions

travankor:
Pending https://github.com/void-linux/void-packages/pull/29434


anatol:
Thanks a lot for the contribution!

PR #131, opened 2022-02-16
image: add workaround to deal with libraries located in /lib

nmeum:
While creating a booster package for Alpine Linux, I noticed that
booster assumes unconditionally that shared libraries are located in
/usr/lib. Contrary to Arch Linux, Alpine uses both /usr/lib and /lib.
For example, the shared library for our musl libc is located in /lib.

This patch adds a workaround which checks /lib (in addition to /usr/lib)
for shared libraries, if the library wasn't found in the latter. A proper
solution would probably be not hardcoding any library directories at all
but this should be fine for now until the comment in `AppendElfDependencies`
is addressed:

https://github.com/anatol/booster/blob/82f015198b46c08f4216152044487760260edbd7/generator/image.go#L324-L325

### DIFF
--- a/generator/image.go
+++ b/generator/image.go
@@ -338,9 +338,18 @@ func (img *Image) AppendElfDependencies(ef *elf.File) error {
 		libs = append(libs, interp)
 	}
 
-	for _, p := range libs {
-		if !filepath.IsAbs(p) {
-			p = filepath.Join("/usr/lib", p)
+	for _, lib := range libs {
+		var p string
+		if filepath.IsAbs(lib) {
+			p = lib
+		} else {
+			p = filepath.Join("/usr/lib", lib)
+
+			// XXX: Workaround to support libraries located in /lib.
+			_, err := os.Open(p)
+			if os.IsNotExist(err) {
+				p = filepath.Join("/lib", lib)
+			}
 		}
 		err := img.AppendFile(p)
 		if err != nil {


anatol:
Your change looks good to me. And I agree that resolving TODO would be a better solution long-term.

nmeum:
The busybox emergency shell feature also seems to assume that busybox is in /usr/bin while it is in /bin on Alpine.

anatol:
booster [adds `bin->/usr/bin` symlink](https://github.com/anatol/booster/blob/275cc76724247812c5ed523a582d9fdb6eae8d71/generator/generator.go#L216) to the image. So this usecase should be covered and using `/usr/bin/busybox` path should not be a problem.

nmeum:
Yeah, booster adds a `/bin → /usr/bin` but it expect busybox to be in `/usr/bin` so what I would need is a symlink from `/usr/bin → to /bin` (if my BusyBox in the image is in `/bin/busybox`):

```
$ booster ls /boot/initramfs-booster | grep bin
usr/sbin -> /usr/bin
bin -> /usr/bin
sbin -> /usr/bin
usr/local/bin -> /usr/bin
usr/local/sbin -> /usr/bin
bin/busybox
```

I don't have `/usr/bin/busybox` with this setup, only `/bin/busybox`, thus the emergency shell doesn't work:

https://github.com/anatol/booster/blob/599c1d2ca2f960aaf81d456474b2c6619710c85c/init/main.go#L794-L800

nmeum:
I am wondering if it makes sense to enforce a canonical location for libraries and binaries in the image, e.g. enforce the use of `/usr/lib` and `/usr/bin` in the image. So instead of installing libraries located on the host in `/lib` to the same directory in the image instead install them to `/usr/lib` (same with the binaries).

This should also resolve an issue I am having with unpack currently. Consider the following image:

```
$ booster ls /boot/initramfs-booster
lib -> /usr/lib
usr/lib/
usr/lib/libc.musl-x86_64.so.1 -> ld-musl-x86_64.so.1
usr/lib/ld-musl-x86_64.so.1
lib/ld-musl-x86_64.so.1
```

If I try to unpack this I get the following error:

```
$ ./booster unpack /boot/initramfs-booster /tmp/unpacked
2022/02/17 19:55:21 open /tmp/unpacked/lib/ld-musl-x86_64.so.1: permission denied
```

This is because `/tmp/unpacked/lib` is a link to `/usr/lib`, i.e. unpack attempts to write `/usr/lib/ld-musl-x86_64.so.1` (to unpack `lib/ld-musl-x86_64.so.1`) which obviously fails without superuser rights. Alternatively, removing the symlinks and having `/lib` and `/usr/lib` be different directories in the image should also work (might even be the cleaner solution).

PR #132, opened 2022-02-16
generator: re-use imageModulesDir constant

nmeum:
At Alpine Linux, Linux Kernel files are also located in /lib instead
of /usr/lib. For this reason, I patch the constants in config.go
accordingly for the Alpine Linux booster package. However, while working
on this I noticed that config.go does not re-use the imageModulesDir
constant. I think it makes sense to re-use it to not hardcode the path
twice.


anatol:
Thank you for the fix!

PR #134, opened 2022-02-17
generator: when creating compatibility symlinks ensure target exists

nmeum:
Otherwise, the symlinks may point to non-existend target directories
if we only install files from /bin and /lib. For example, on Alpine
BusyBox is located in /bin/busybox and hence the /usr/bin directory
is not created without this commit.


anatol:
Thank you @nmeum!

PR #139, opened 2022-03-12
all: fix a few staticcheck warnings

mvdan:
(see commit message)


generator/console.go
```diff
@@ -136,7 +136,6 @@ func readFontFile(font string) (blob []byte, err error) {
 				if err != nil {
 					return nil, err
 				}
-				name = name[:len(name)-3] // remove .gz suffix from the name
```
anatol, line 139: thanks for catching it.

it is a leftover from previous refactoring

PR #143, opened 2022-04-02
generator: lookup executable paths using exec.LookPath

nmeum:
This fixes an issues I discovered on Alpine when attempting to use Booster's vconsole feature as the setfont executable is located in /usr/sbin on Alpine and hence the binary was not found successfully by the primitive lookup implementation previously provided by Booster. I think there should be no downside to using [`exec.Lookup`](https://pkg.go.dev/os/exec@go1.18#LookPath) hence proposing this for upstream inclusion. While at it, I also made use of [`filepath.IsAbs`](https://pkg.go.dev/path/filepath#IsAbs).

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -256,6 +256,11 @@ func (img *Image) appendExtraFiles(binaries []string) error {
 			if err != nil {
 				return err
 			}
+
+			if err := img.AppendFileTo(f, "/usr/bin"); err != nil {
+				return err
+			}
+			continue
 		}
 
 		if err := img.AppendFile(f); err != nil {
--- a/generator/image.go
+++ b/generator/image.go
@@ -200,10 +200,14 @@ func (img *Image) AppendContent(dest string, osMode os.FileMode, content []byte)
 	return img.AppendEntry(dest, mode, content)
 }
 
-// AppendFile appends the file + its dependencies to the ramfs file
+// AppendFileTo appends the file + its dependencies to the ramfs file at the given destination.
 // If input is a directory then content is added to the image recursively.
-func (img *Image) AppendFile(fn string) error {
+func (img *Image) AppendFileTo(fn string, destDir string) error {
 	fn = path.Clean(fn)
+	destDir = path.Clean(destDir)
+
+	// Destination path in ramfs file (according to destDir).
+	destFilePath := path.Join(destDir, path.Base(fn))
 
 	img.m.Lock()
 	if img.contains[fn] {
@@ -212,7 +216,7 @@ func (img *Image) AppendFile(fn string) error {
 	}
 	img.m.Unlock()
 
-	if err := img.AppendDirEntry(path.Dir(fn)); err != nil {
+	if err := img.AppendDirEntry(destDir); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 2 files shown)


generator/generator.go
```diff
@@ -247,9 +248,14 @@ func (img *Image) appendInitBinary(initBinary string) error {
...
-			// simple names like "strace" are resolved as binaries under /usr/bin
-			f = "/usr/bin/" + f
+		if !filepath.IsAbs(f) {
+			// If the given name is not an absolute path, assume that it refers
+			// to an executable and lookup the path to the executable using $PATH.
+			var err error
+			f, err = exec.LookPath(f)
+			if err != nil {
+				return err
+			}
```
nmeum, line 258: I think we need to make sure that the file is still added to /usr/bin afterwards, even if originates it a different location on the host.

anatol: This is one option indeed.

Another option is to add binaries with `/usr/sbin/` path and then modify `PATH` envvar in ` boost()` function.

anatol:
The changes themselves (exec.Lookup, isAbs) look good to me.

generator/generator.go
nmeum: I have added an `img.AppendFileTo` function and started using this for adding binaries. Haven't tested this extensively yet, but at least it seems to work for `setfont`. Let me know what you think.

generator/generator.go
```diff
@@ -247,9 +248,18 @@ func (img *Image) appendInitBinary(initBinary string) error {
...
+
+			if err := img.AppendFileTo(f, "/usr/bin"); err != nil {
```
anatol, line 260: you have 2 `AppendFile()` for the same file `f` inside the loop. I think it should be called only once.

nmeum: Good catch, fixed.

anatol:
I pushed two first commits to `master`.

As for the "move all binaries to /usr" idea I need to think more. One concern I have is that `exec.LookPath` uses `$PATH` to lookup binaries. If a malicious location is added somehow to user's $PATH then the binary overrides the system location. I would much prefer that booster used binaries from the system path only and added to a well-known location at the booster image.

@nmeum to understand your situation better, could you please pull the latest `master` and try to reproduce the problem with `setfont` you have. Build the image and show me output of `booster ls /boot/your-booster.img`. And please post output of the booster boot logs.

nmeum:
> As for the "move all binaries to /usr" idea I need to think more. One concern I have is that `exec.LookPath` uses `$PATH` to lookup binaries. If a malicious location is added somehow to user's $PATH then the binary overrides the system location. I would much prefer that booster used binaries from the system path only and added to a well-known location at the booster image.

That is fiine with me.

> @nmeum to understand your situation better, could you please pull the latest `master` and try to reproduce the problem with `setfont` you have. Build the image and show me output of `booster ls /boot/your-booster.img`. And please post output of the booster boot logs.

I didn't have a particular problem with `setfont`. My thinking was just that if we lookup binaries (like `setfont`) using `$PATH` (i.e. via `exec.LookPath`) we might end up finding a binary on the host which is not in `/usr/bin:/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin` (i.e. would not end up in the `$PATH` of the generated image). For example, I occasionally have stuff installed to `~/bin` for testing purposes. In this case, booster would add `~/bin/setfont` to the image in a location that wouldn't be picked up by booster. I was jut thinking that adding binaries to `/usr/bin` in the image would be better from a reliability/correctness perspective but since it is probably rare for binary like `setfont` to appear outside the aforementioned "standard directories" I am also complete fine with ignoring this problem for now :)

anatol:
> I am also complete fine with ignoring this edge case for now :)

Ok, let's ignore it for now.

Could you please confirm that `setfont` is detected and used properly at Alpine now?

nmeum:
Yes, I can confirm that it works on Alpine :rocket:

anatol:
Nice! Thank you for your patches and your feedback!

Let me know if you see other issues at Alpine.

PR #145, opened 2022-04-03
generator: Don't error-out if /etc/locale.conf doesn't exist

nmeum:
musl-based Linux distributions (e.g. Alpine Linux) don't use `/etc/locale.conf` since musl supports UTF-8 by default and doesn't require any external locale files. To improve support for musl-based Linux-distributions, this patches causes Booster to not error-out when `/etc/locale.conf` doesn't exist. Furthermore, if no language is defined in /etc/locale.conf (or the file doesn't exist) UTF-8 support is assumed.

I am unsure if this breaks anything on glibc-based systems. Proposing this for upstream inclusion to discuss this further. An alternative would be to just allow `/etc/locale.conf` to not exist but don't assume UTF-8 support (if no LANG is defined).

### DIFF
--- a/generator/console.go
+++ b/generator/console.go
@@ -3,8 +3,10 @@ package main
 import (
 	"bytes"
 	"compress/gzip"
+	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path"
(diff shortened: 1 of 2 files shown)


anatol:
This also needs an integration test. I will look at it early next week.

anatol:
Hi @nmeum

I cleaned up the logic a bit and added tests for your change. Please try it and let me know if it works properly with Alpine Linux.

nmeum:
> I cleaned up the logic a bit and added tests for your change. Please try it and let me know if it works properly with Alpine Linux.

Thanks a lot! I can confirm that it still works properly on Alpine :tada:

Issue #144, opened 2022-04-03
Local echo disabled when emergency shell is entered after mount timeout during password prompt


nmeum:
While performing some tests with Booster on LUKS-encrypted system. I noticed that the BusyBox emergency shell does not have local echo enabled when it is started due to the `mount_timeout` option while Booster is prompting for a password.

This is due to the fact that the `readPassword` function correctly disables local echo:

https://github.com/anatol/booster/blob/8666e7b36674976ea83093ad5c2b83b0528617c7/init/console.go#L195-L201

The function also restores local echo via `defer`:

https://github.com/anatol/booster/blob/8666e7b36674976ea83093ad5c2b83b0528617c7/init/console.go#L203

However, I believe on mount timeout this function will not return (since it is executed in a different goroutine) and thus the defer statement is not executed. Hence, I think the `emergencyShell` implementation needs to restore some sort of sane stty settings:

https://github.com/anatol/booster/blob/edeff4c705d5f2b3a02162bed74626b70754ca63/init/main.go#L790-L796

This could either be achieved by executing something like `busybox stty sane` before spawing the shell (which would require the stty Busybox applet) or, alternatively, by manually modifying the termios settings accordingly (see `readPassword` code above).

PR #146, opened 2022-04-07
Unconditionally enable local echo for emergency shell

nmeum:
Fixes https://github.com/anatol/booster/issues/144

### DIFF
--- a/init/main.go
+++ b/init/main.go
@@ -787,8 +787,28 @@ func readStartTime() {
 	}
 }
 
+func enableLocalEcho() error {
+	stdin := os.Stdin
+	fd := int(stdin.Fd())
+
+	termios, err := unix.IoctlGetTermios(fd, unix.TCGETS)
+	if err != nil {
+		return err
+	}
+
+	newState := *termios
+	newState.Lflag |= unix.ECHO
+
+	return unix.IoctlSetTermios(fd, unix.TCSETS, &newState)
+}
+
 func emergencyShell() {
 	if _, err := os.Stat("/usr/bin/busybox"); !os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


init/main.go
```diff
@@ -787,7 +787,30 @@ func readStartTime() {
...
+	if err := unix.IoctlSetTermios(fd, unix.TCSETS, &newState); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func emergencyShell() {
+	// Force local echo (might have been disabled by readPassword).
+	if err := enableLocalEcho(); err != nil {
```
anatol, line 810: let's move the `enableLocalEcho` call under `if _, err := os.Stat("/usr/bin/busybox"); !os.IsNotExist(err) {` so the echo will be disable only if it is needed.

anatol (commented):
the rest looks good to me

init/main.go
```diff
@@ -787,7 +787,30 @@ func readStartTime() {
...
+	fd := int(stdin.Fd())
+
+	termios, err := unix.IoctlGetTermios(fd, unix.TCGETS)
+	if err != nil {
+		return err
+	}
+
+	newState := *termios
+	newState.Lflag |= unix.ECHO
+	if err := unix.IoctlSetTermios(fd, unix.TCSETS, &newState); err != nil {
```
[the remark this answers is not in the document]
nmeum: Right, I just wasn't sure if this style is preferred. Adjusted accordingly, thank you!

init/main.go
nmeum: Good catch, fixed.

anatol:
Thank you!

PR #156, opened 2022-06-11
Always use path/filepath pkg for operating systems paths

nmeum:
The `path` pkg is only intended to be used for “paths separated by forward slashes”. Even though Booster only supports Linux and Linux only uses slash-separated paths I think it is still cleaner to always use `path/filepath` over `path` for operating system file paths. As such, this is only a very small refactoring and (hopefully) shouldn't cause any functional changes.

I only noticed this because some files (e.g. `generator/image.go`) import both `path` as well as `path/filepath` but seem to use path for operating system paths.

### DIFF
--- a/generator/config.go
+++ b/generator/config.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"path"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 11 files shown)


nmeum:
BTW: Speaking of minor refactorings, is there any reason why the configuration file doesn't use YAML Lists but instead relies on comma-separated strings for configuration keys like `modules_force_load`?

anatol (approved):

anatol:
> BTW: Speaking of minor refactorings, is there any reason why the Booster configuration file doesn't use YAML Lists but instead relies on comma-separated strings for configuration keys like `modules_force_load`?

The reason (which is rather weak) was to make the modules list more dense and readable. A standard YAML list would work here as well.

PR #157, opened 2022-06-11
generator: Convert constants pointing to directory to variables

nmeum:
This allows modifying their value via the `-X` flag provided by `go tool link`. This eases packaging of Booster as distributions which use different paths for these directories don't need to patch Booster but instead can set these variables using:

	go build --ldflags="-X main.consolefontsDir=/foo/bar/…"


anatol:
Thank you for the patches @nmeum!

Issue #73, opened 2021-03-28
Advanced parser for kernel boot params


anatol:
Currently booster reads `proc/cmdline` and splits it by space. It works for majority of simple cases but it will fail if something more advanced is used in the command line.

For example parameters itself may use space. In this case quoting around the params need to be used. Per [kernel documentation](https://www.kernel.org/doc/html/v5.11/admin-guide/kernel-parameters.html):
> Double-quotes can be used to protect spaces in values

But it also looks like some bootloaders (e.g. grub) [add quoting in a different way](https://bugzilla.redhat.com/show_bug.cgi?id=950760), like this `"param=value with a space"`. So booster needs to handle quoting in both ways: `param="value with a space"` and `"param=value with a space"`. This in addition to quotes around UUID handled here #6.

Besides spaces the values might contain escape symbols `param="foo \" bar"`. And parser probably needs to handle other param delimiter, e.g. params might be separated with tabs.

It looks like we need a real advanced parser that handles all these cases. It worth reviewing kmod cmdline parser here to see if we can go similar way https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/tree/libkmod/libkmod-config.c?id=8742be0aa5ed14d28980eb706b08e8bdf650f21c#n528

anatol:
`TestLUKS2WithQuotesOverUUID` test is failing now with

```
[    0.807449] booster: while parsing cmdline parameter unexpected " found at 98, input may be malformed, attempting to proceed
[    0.808704] booster: while parsing cmdline parameter unexpected " found at 135, input may be malformed, attempting to proceed
[    0.809946] booster: root=UUID="7bbf9363-eb42-4476-8c1c-9f1f4d091385": unable to parse UUID parameter UUID="7bbf9363-eb42-4476-8c1c-9f1f4d091385": expected input length is 36, got length 38
[    0.811831] booster: Press ENTER to reboot
```

jacobmyers-codeninja:
I know whats causing this, I should have a small fix to address it and a few tests cases to ensure its covered better.

PR #160, opened 2022-07-16
ISSUE-73: Improved parser for cmdline parameters

jacobmyers-codeninja:
For #73 - Adding a more advanced parser to allow handling more complex cmdline parameters. This is based more on a 1 rune parse state machine style design, it iterates over each rune and makes a decision on what to do, copy/exit/change modes. Tried to leave enough comments to make it less confusing.

- This can split apart on any whitespace (null bytes, newlines, carriage returns,tabs, spaces)
- Quotes are handled and stripped from the strings
- Escaping quotes (or other chars) will copy them through
- Keys can't have an =, but values can get = chars in
- Whitespace is generally cleaned up

The design is around a single key/value enumeration approach, so it can be called until everything is processed. It handles some bad conditions ok, other things it might choke on. Worst case it outputs a bad chunk of data as the final result if the input is horribly out of spec/mangled. It generates new strings for each key/value it reads, using += assembly so not the best, but its a one shot deal so shouldn't be a huge deal.

I did not clean up some of the other code that could probably be removed by using this (such as the strip quotes in luks uuid, I believe this will allow that to be avoided as the quotes are removed through this.

I used the following string as test:
```
"    key=value   param=test=true param=\"test=true\"   test=tr\\\000ue double=equals=true \"param=tr ue\" \"param=yes\" param2=\"also yes\" quotes=\"escaped\\\"ooh\" ro root=/dev/mapper/VolGroup-lv_root rd_NO_LUKS KEYBOARDTYPE=pc KEYTABLE=us\nLANG=en_US.UTF-8 rd_NO_MD rd_LVM_LV=VolGroup/lv_swap SYSFONT=latarcyrheb-sun16\nrd_LVM_LV=VolGroup/lv_root rd_NO_DM rhgb quiet selinux=0"
```

Results: [ next token start, 'key parsed', 'value parsed' ]
```
        14      'key' = 'value'
        32      'param' = 'test=true'
        50      'param' = 'test=true'
        64      'test' = 'true'
        83      'double' = 'equals=true'
        97      'param' = 'tr ue'
        109     'param' = 'yes'
        127     'param2' = 'also yes'
        149     'quotes' = 'escaped"ooh'
        152     'ro' = ''
        186     'root' = '/dev/mapper/VolGroup-lv_root'
        197     'rd_NO_LUKS' = ''
        213     'KEYBOARDTYPE' = 'pc'
        225     'KEYTABLE' = 'us'
        242     'LANG' = 'en_US.UTF-8'
        251     'rd_NO_MD' = ''
        278     'rd_LVM_LV' = 'VolGroup/lv_swap'
        304     'SYSFONT' = 'latarcyrheb-sun16'
        340     'rd_NO_DM' = ''VolGroup/lv_root'
        345     'rhgb' = ''
        351     'quiet' = ''
        360     'selinux' = '0'
```

### DIFF
--- a/init/cmdline.go
+++ b/init/cmdline.go
@@ -83,7 +83,7 @@ func getNextParam(params string, index int) (string, string, int) {
 		}
 
 		switch r {
-		case '\000', '\n', '\r', '\t', ' ':
+		case 0, '\n', '\r', '\t', ' ':
 			// if we haven't seen any non-whitespace yet just continue
 			if !copyMode {
 				continue
@@ -101,10 +101,31 @@ func getNextParam(params string, index int) (string, string, int) {
 			// escaping something, update flag and move on
 			escaping = true
 		case '"':
-			// we changed quote mode
-			inQuote = !inQuote
-			// also now in copy mode if we were not already
+			// now in copy mode if we were not already
 			copyMode = true
+
+			// if we are in quote mode this ends it
+			if inQuote {
+				inQuote = false
+
+				// if we have parsed a key already this ends our parse too, otherwise continue as normal
+				if keyComplete {
+					return key, value, index + i + 1
+				}
+
+				continue
+			}
+
+			// if we are parsing a key, and it isn't empty, then something has gone wrong
+			// same for value
+			if (!keyComplete && len(key) > 0) || (keyComplete && len(value) > 0) {
+				// error, this quote is inside real characters
+				// we are going to recover as best we can, just copy the quote and hope for the best
+				copyRune(r)
+				continue
+			}
+
+			inQuote = true
 		case '=':
 			// this separates key=value, but only while in key mode
 			if !keyComplete {
@@ -196,7 +217,7 @@ func parseParams(params string) error {
 			if len(parts) != 2 {
 				return fmt.Errorf("invalid rd.luks.name kernel parameter %s, expected format rd.luks.name=<UUID>=<name>", value)
 			}
-			uuid, err := parseUUID(stripQuotes(parts[0]))
+			uuid, err := parseUUID(parts[0])
 			if err != nil {
 				return fmt.Errorf("invalid UUID %s %v", parts[0], err)
 			}
@@ -207,15 +228,14 @@ func parseParams(params string) error {
 			}
 			luksMappings = append(luksMappings, dev)
 		case "rd.luks.uuid":
-			stripped := stripQuotes(value)
-			u, err := parseUUID(stripped)
+			u, err := parseUUID(value)
 			if err != nil {
 				return fmt.Errorf("invalid UUID %s in rd.luks.uuid boot param: %v", value, err)
 			}
 
 			dev := luksMapping{
 				ref:  &deviceRef{refFsUUID, u},
-				name: "luks-" + stripped,
+				name: "luks-" + value,
 			}
 			luksMappings = append(luksMappings, dev)
 		default:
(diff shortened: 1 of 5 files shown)


anatol:
Thank you for this change! It looks great.

> This can split apart on any whitespace (null bytes, newlines, carriage returns,tabs, spaces)
Quotes are handled and stripped from the strings
Escaping quotes (or other chars) will copy them through
Keys can't have an =, but values can get = chars in
Whitespace is generally cleaned up

Could you please add unit tests for these ^^ cases?

>  Worst case it outputs a bad chunk of data as the final result if the input is horribly out of spec/mangled.

And for this as well. Basically, we want as full unit test coverage as possible to handle future code changes smoother.

jacobmyers-codeninja:
I added test cases to cover most examples of how the parser works, it has no error handling other than to try its best to recover. Garbage in, garbage out, wasn't sure if there would be a desire to throw any errors or not, could probably at least add warnings that unexpected characters were encountered perhaps.

anatol:
> could probably at least add warnings that unexpected characters were encountered perhaps.

Yeah, I think it is a good idea. If the command line does not parse correctly, then post a `warning()`. The message will go to the kernel logs, and it will help the user figure out what is going on with the system.

init/cmdline.go
```diff
@@ -122,7 +219,7 @@ func parseParams(params string) error {
 			if len(parts) != 2 {
 				return fmt.Errorf("invalid rd.luks.name kernel parameter %s, expected format rd.luks.name=<UUID>=<name>", value)
 			}
-			uuid, err := parseUUID(stripQuotes(parts[0]))
```
anatol, line 125: Can the `stripQuotes()` function be removed now?

jacobmyers-codeninja: It does look like it, everything it was used for came from the cmdline parser and it cleans them up automatically now

anatol:
I see that your commits have author set to `32110416+jacobmyers-codeninja@users.noreply.github.com` Do you want to set some real email instead?

anatol:
One more suggestion is to use `gofumpt -l -w .` formatter that applies stricter formatting rules and used by booster. I ran this formatter locally for your change but it is a good idea to use it for your changes in the future.

anatol:
Otherwise the change looks good and I would like to apply it to `master`. @jacobmyers-codeninja do you have anything to add to this PR?

jacobmyers-codeninja:
> I see that your commits have author set to `32110416+jacobmyers-codeninja@users.noreply.github.com` Do you want to set some real email instead?

Unfortunately I already get way too much junk to my accounts for email to even be a good contact option, so I just use the github standin. It links to my github account which people are free to contact me though and allows me to easily keep the github alert mails sorted.

> One more suggestion is to use `gofumpt -l -w .` formatter that applies stricter formatting rules and used by booster. I ran this formatter locally for your change but it is a good idea to use it for your changes in the future.

I forgot to run code cleanup, but it looks like it was just removing the redundant test{} declarations since the type info was known in advance. I only have my windows machine available for a while, when I try to run the go formatter it changes all the line endings and produces a commit with every file changed :(

> Otherwise the change looks good and I would like to apply it to `master`. @jacobmyers-codeninja do you have anything to add to this PR?

I don't see any immediate cause for concern with anything, its as well put together as my primitive go understanding allows :) - I am intending to eventually use this for some servers I am migrating to void when I get everything relocated and running and I need luks keyfile support to make it happen. I figured this looked like something I could take a stab at to get acquainted with to experiment with that PR more as well.

I should have actual equipment to do more testing with in a few weeks.

jacobmyers-codeninja:
I managed to get git and gofmt to behave with each other so ran it through gofmt! \o/

anatol:
> I managed to get git and gofmt to behave with each other so ran it through gofmt! \o/

I suggest you to check https://github.com/mvdan/gofumpt - it is a stricter version of gofmt from other of the go developers. booster uses it for formatting.

anatol:
This PR has been squashed and merged `master` as fad77ecb30fa7fa37848748e6a93c6a1746fd912. Thank you for your work @jacobmyers-codeninja !

Issue #37, opened 2021-02-24
Add support for unlocking LUKS with a key file


joserebelo:
It would be nice if something like the encrypt hook `cryptkey` or sd-encrypt's `rd.luks.key` was supported, allowing for the root partition to be unlocked with a key file.

More info: https://wiki.archlinux.org/index.php/Dm-crypt/System_configuration#rd.luks.key

gadgethm:
I also would like this support, as I use a TPM to automatically unlock my encrypted drive.

dkwo:
Just a quick question: is this now implemented? if so, does it mean one can use yubikey with booster? or is it in progress?
Thanks.

anatol:
@dkwo the feature is implemented now. See https://github.com/jacobmyers-codeninja/booster/commit/ab56ce57bb5ee246f910f4d381b15f3a25971c59#diff-202700f9afad82959d12d36af398c56ad7a0e5223da8382e5a4aced3253da171

Note that the keyfile must live at the initramfs.

Barbaross93:
If we want to include the key file within the initramfs, then would we want to include it with the `extra_files: ...` option?

Barbaross93:
Nevermind, that does seem to be the way to do it. Unfortunately, this doesn't seem to be very secure for unlocking the root partition. Support for specifying a key file on a separate disk, e.g. a USB stick would be ideal. Are/is there any plans/interest in expanding on this feature?

Alternatively, and I can make a separate issue for this, how does the clevis style binding of LUKS work with booster? I'm running on Void Linux which doesn't have systemd or clevis currently in the repos. There are some tpm2 utilities in the repos, so a manual way of binding LUKS with the tpm could be possible for my use case. I'm just unsure of what  commands need to be run.

anatol:
> Support for specifying a key file on a separate disk, e.g. a USB stick would be ideal.

Yes, the plan is to support following features:
 1. reading the unlock key from a file at $EFI
 2. (optionally) read key from a block device by its offset and key length (e.g. "read key from UUID=283983423 offset=4096 length=1024")

> how does the clevis style binding of LUKS work with booster?

Booster has a great support for Clevis style binding (Tang/TPM2/SSS) plus it supports custom Yubikey Challenge-Response binding.

In addition to that Booster supports Systemd TPM Fido style binding.

> I'm just unsure of what commands need to be run.

The best option is to add `clevis` and `tang` packages to Void repo. And then use the standard enrollment operations.

[...]

PR #161, opened 2022-07-17
ISSUE-37: Allow luks unlocking via keyfile specified in kernel parameters with rd.luks.key

jacobmyers-codeninja:
For #37: Add support to allow unlocking a luks volume by keyfile, reworked the luks mappings a little to allow more complex patterns of parsing parameters and building mappings.

This slightly tweaks creating luksMapping lists so they can occur multiple times in the kernel parameter list and it just updates the relevant properties for each mapping (to allow extending even further if desired).

- updated luksMapping type to hold a keyfile
- add a luksMapping findOrCreate function in luks.go, either finds the existing mapping by UUID or adds one
- update the rd.luks.uuid and rd.luks.name in cmdline.go to make use of findOrCreate 
- adds rd.luks.key with the format UUID=keyfile which should point to a file in initramfs containing the password
- adds recoverKeyfilePassword go routine which attempts to read the password from the keyfile and unlock the device, falling back to running the requestKeyboardPassword option if that fails
- update luksOpen to call recoverKeyfilePassword if a keyfile is defined on the mapping

Putting this forward more as an idea for the approach, not sure how good or bad the specific way I implemented it is.

### DIFF
--- a/init/cmdline.go
+++ b/init/cmdline.go
@@ -220,27 +220,52 @@ func parseParams(params string) error {
 			if len(parts) != 2 {
 				return fmt.Errorf("invalid rd.luks.name kernel parameter %s, expected format rd.luks.name=<UUID>=<name>", value)
 			}
+
 			uuid, err := parseUUID(parts[0])
 			if err != nil {
 				return fmt.Errorf("invalid UUID %s %v", parts[0], err)
 			}
 
-			dev := luksMapping{
-				ref:  &deviceRef{refFsUUID, uuid},
-				name: parts[1],
-			}
-			luksMappings = append(luksMappings, dev)
+			m := findOrCreateLuksMapping(uuid)
+			m.name = parts[1]
 		case "rd.luks.uuid":
-			u, err := parseUUID(value)
+			uuid, err := parseUUID(value)
 			if err != nil {
 				return fmt.Errorf("invalid UUID %s in rd.luks.uuid boot param: %v", value, err)
 			}
 
-			dev := luksMapping{
-				ref:  &deviceRef{refFsUUID, u},
-				name: "luks-" + value,
+			findOrCreateLuksMapping(uuid)
+		case "rd.luks.key":
+			var uuid UUID
+			var keyfile string
+
+			parts := strings.SplitN(value, "=", 2)
+
+			if len(parts) == 1 {
+				// do we only have 1 luks device?
+				if len(luksMappings) == 1 {
+					// we attach to it and hope for the best
+					uuid = luksMappings[0].ref.data.(UUID)
+				} else {
+					// don't know what to do here
+					return fmt.Errorf("invalid rd.luks.key kernel parameter %s, more than 1 luks device", value)
+				}
+
+				keyfile = parts[0]
+			} else if len(parts) == 2 {
+				var err error
+				uuid, err = parseUUID(parts[0])
+				if err != nil {
+					return fmt.Errorf("invalid UUID %s in rd.luks.key boot param: %v", value, err)
+				}
+
+				keyfile = parts[1]
+			} else {
+				return fmt.Errorf("invalid rd.luks.key kernel parameter %s, expected format rd.luks.key=<UUID>=<keyfile>", value)
 			}
-			luksMappings = append(luksMappings, dev)
+
+			m := findOrCreateLuksMapping(uuid)
+			m.keyfile = keyfile
 		case "zfs":
 			zfsDataset = value
 		default:
--- a/init/luks.go
+++ b/init/luks.go
@@ -12,6 +12,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"regexp"
 	"strings"
 	"time"
 
@@ -24,6 +25,7 @@ import (
 type luksMapping struct {
 	ref     *deviceRef
 	name    string
+	keyfile string
 	options []string
 }
 
@@ -298,6 +300,47 @@ func recoverTokenPassword(volumes chan *luks.Volume, d luks.Device, t luks.Token
 	info("password from %s token #%d does not match", t.Type, t.ID)
 }
 
+func recoverKeyfilePassword(volumes chan *luks.Volume, d luks.Device, checkSlots []int, mappingName string, keyfile string) {
+	var err error
+	var password []byte
+
+	// keyfile might be in the format /path:UUID=<DEV UUID> to indicate the keyfile lives on another device
+	parts := regexp.MustCompile("(?i):UUID=").Split(keyfile, 2)
+
+	if len(parts) == 1 {
+		password, err = os.ReadFile(parts[0])
+
+		if err != nil {
+			warning("reading password: %v", err)
+		}
+	} else {
+		// read password from device matching uuid
+		uuid, err := parseUUID(parts[1])
+		if err != nil {
+			warning("invalid UUID %s in rd.luks.key boot param: %s", uuid, keyfile)
+		} else {
+			// TODO: access path on uuid device, read password
+			warning("user wants keyfile from device %s, but I don't know how to do that", uuid)
+		}
+	}
+
+	if len(password) > 0 {
+		for _, s := range checkSlots {
+			v, err := d.UnsealVolume(s, password)
+			if err == luks.ErrPassphraseDoesNotMatch {
+				continue
+			}
+			volumes <- v
+			return
+		}
+	}
+
+	warning("password in keyfile #{keyfile} was unable to unseal #{mappingName}\n")
+
+	// have to use keyboard password
+	requestKeyboardPassword(volumes, d, checkSlots, mappingName)
+}
+
 func requestKeyboardPassword(volumes chan *luks.Volume, d luks.Device, checkSlots []int, mappingName string) {
 	for {
 		prompt := fmt.Sprintf("Enter passphrase for %s:", mappingName)
@@ -369,7 +412,13 @@ func luksOpen(dev string, mapping *luksMapping) error {
 		}
 	}
 	if len(checkSlotsWithPassword) > 0 {
-		go requestKeyboardPassword(volumes, d, checkSlotsWithPassword, mapping.name)
+		// is there a keyfile defined for the password for this volume?
+		if len(mapping.keyfile) > 0 {
+			// if the keyfile doesn't work we will fallback to password
+			go recoverKeyfilePassword(volumes, d, checkSlotsWithPassword, mapping.name, mapping.keyfile)
+		} else {
+			go requestKeyboardPassword(volumes, d, checkSlotsWithPassword, mapping.name)
+		}
 	}
 
 	v := <-volumes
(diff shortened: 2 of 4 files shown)


init/luks.go
```diff
@@ -411,3 +440,23 @@ func handleLuksBlockDevice(blk *blkInfo) error {
...
+	}
+
+	for _, o := range luksMappings {
+		if blk.matchesRef(o.ref) {
+			return &o
+		}
+	}
+
+	// didn't locate the device make a new one
+	luksMappings = append(luksMappings, luksMapping{
```
anatol, line 456: this part can be simpler - make a variable for the new luksMapping and then return it
```go
m := luksMapping{
                ref:  &deviceRef{refFsUUID, uuid},
		name: "luks-" + uuid.toString(),
	}
luksMappings = append(luksMappings, m)
return &m
```

init/cmdline.go
```diff
@@ -127,23 +127,29 @@ func parseParams(params string) error {
...
-				ref:  &deviceRef{refFsUUID, uuid},
-				name: parts[1],
-			}
-			luksMappings = append(luksMappings, dev)
+			var o = findOrCreateLuksMapping(uuid)
+			o.name = parts[1]
 		case "rd.luks.uuid":
 			stripped := stripQuotes(value)
-			u, err := parseUUID(stripped)
+			uuid, err := parseUUID(stripped)
```
anatol, line 134: this part needs to be rebased on top of `master` that contains your other changes

init/luks.go
```diff
@@ -299,6 +301,27 @@ func recoverTokenPassword(volumes chan *luks.Volume, d luks.Device, t luks.Token
 	info("password from %s token #%d does not match", t.Type, t.ID)
 }
 
+func recoverKeyfilePassword(volumes chan *luks.Volume, d luks.Device, checkSlots []int, mappingName string, keyfile string) {
+	password, err := ioutil.ReadFile(keyfile)
```
anatol, line 305: does it mean you are trying to read the password from the file packaged into the initramfs? Is it normal practice?

I would always assume a standard place to store the password file is either EFI partition (next to kernel and initramfs) or an external USB flash device.

jacobmyers-codeninja: Yes, a fairly normal practice is to store the password within initramfs which is itself stored in /boot, which itself is encrypted and requires grub (or your bootloader of choice) to unlock. Either because root contains /boot, or in many current cases a separate luksv1 partition (because grub and luksv2 don't really play nice still). The reason for this being a normal approach is it is often the "easiest" since init can easily access files in initramfs.

The file could also be stored in any other encrypted location, assuming it has been unlocked already. The primary use case for the keyfile is to prevent needing to type a password multiple times (once for grub, once for luks to unseal) when using full disk encryption. It can be combined with TPM autounsealing though to potentially allow unattended boot of encrypted volumes that contain the additional keys to unseal the remaining volumes too, but I view that as a security risk personally and require a password entry on every boot to unseal the boot partition. In these cases EFI will contain a single efi executable, the signed grub with all its modules preloaded and the bundled config pointing it to whatever volume holds /boot.

Storing it on an external drive is also an option, but that will require implementing the full keyfile spec to allow pointing to devices as well. I can certainly take a look at how complex that process would be to implement. I wouldn't mind storing the key right on /boot and referencing it directly, I was just starting with the lowest hanging fruit first :)

The primary reason for all this is that TPM is still a minefield on a lot of platforms (ARM for instance, VMs for another) so there is a lot of cases where people are just using secureboot combined with luks to harden security in those cases. It just also happens that fast and lightweight principals of booster coincide with those platforms use cases.

This gives me some stuff to look at more and once I get my systems up to where I can test I can toy with a few configurations as well, thanks!

init/luks.go
jacobmyers-codeninja: Good catch, I toyed with that function a bit changing what I wanted to do with it and didn't fully clean it up. I threw it in luks since it is really pretty luks specific, but I do see the utils file as well, as this is also mostly a utility function would it be better suited there? Wasn't sure what you preference was on locality.

init/cmdline.go
```diff
@@ -127,23 +127,29 @@ func parseParams(params string) error {
 				return fmt.Errorf("invalid UUID %s %v", parts[0], err)
 			}
 
-			dev := luksMapping{
-				ref:  &deviceRef{refFsUUID, uuid},
-				name: parts[1],
-			}
-			luksMappings = append(luksMappings, dev)
+			var o = findOrCreateLuksMapping(uuid)
```
anatol, line 130: maybe use name `m` (for mapping) here?

init/luks.go
anatol: I have no strong preference, so let's keep it here

init/cmdline.go
```diff
@@ -127,23 +127,29 @@ func parseParams(params string) error {
...
-			u, err := parseUUID(stripped)
+			uuid, err := parseUUID(stripped)
 			if err != nil {
 				return fmt.Errorf("invalid UUID %s in rd.luks.uuid boot param: %v", value, err)
 			}
 
-			dev := luksMapping{
-				ref:  &deviceRef{refFsUUID, u},
-				name: "luks-" + stripped,
+			findOrCreateLuksMapping(uuid)
```
anatol, line 139: It looks like setting name to `name: "luks-" + stripped,` is lost here.

init/luks.go
anatol: Ok fail points.

Though I would love to have a way to keep a keyfile at EFI or and USB drive as well. And we need to think about how to reference different locations. if initramfs-based key referenced by `rd.luks.key=UUID=path` what would be the configuration for EFI based one? for a USB based one?

anatol: Speaking of TPM key wrapping clevis is a good library to do it. We can use it to wrap the keyfile with TPM or Yubikey.

https://github.com/anatol/clevis.go

init/luks.go
```diff
@@ -370,7 +393,13 @@ func luksOpen(dev string, mapping *luksMapping) error {
 		}
 	}
 	if len(checkSlotsWithPassword) > 0 {
-		go requestKeyboardPassword(volumes, d, checkSlotsWithPassword, mapping.name)
+		// is there a keyfile defined for the password for this volume?
+		if len(mapping.keyfile) > 0 {
```
anatol, line 397: If the keyfile is specified do we expect that at least one LUKS slot could be unlocked with it? In this case, do we still want to request password using keyboard?

anatol (commented):
A related note. I think that the current way to specify luks parameters is a bit confusing when we have multiple LUKS partitions.

Here is a change that tries to reorganize luks options and make it more consistent
https://github.com/anatol/booster/commit/43d2c0c31bad6d226d6c2f3cf57f152c1f6828fd

init/luks.go
jacobmyers-codeninja: > if initramfs-based key referenced by `rd.luks.key=UUID=path` what would be the configuration for EFI based one? for a USB based one?

The official keyflie spec is `rd.luks.key=<UNLOCK UUID>=path:UUID=<STORAGE UUID>` where `:UUID=<STORAGE UUID>` is optional. Technically the unlock uuid is also optional, if not specified it is assumed to be the root/old provided rd.luks device specified. The storage uuid is expected to be the uuid of whatever device is holding the keyfile and is assumed you have provided necessary filesystem drivers to open it.

init/cmdline.go
jacobmyers-codeninja: findOrCreate defaults to name = luks-UUID when it creates

init/luks.go
jacobmyers-codeninja: A keyfile is just a file containing a passphrase. Often these are large (4k of random data), but they are done in addition to an actual password in another slot. The keyfile is for unsealing without manual intervention (avoiding the 2nd round of entering a password for instance), however should that keyfile become lost or damaged the original password is still usable. The keyfile and passwords use the same slot types, so in this case I am giving them the opportunity to manually unseal and finish booting using a password if keyfile fails so they can fix the issue (maybe they forgot to add the file to initramfs, lost the thumbdrive holding it/etc) instead of just locking them completely out of the computer.

init/cmdline.go
```diff
@@ -127,23 +127,29 @@ func parseParams(params string) error {
...
+		case "rd.luks.key":
```
anatol, line 140: This parameter needs to be documented in the `docs/`

anatol:
Okay, the change looks fine. A few minor things to do - document the new param, and refactor `findOrCreateLuksMapping` to make it more readable.

init/cmdline.go
jacobmyers-codeninja: Will work on that, also examining what options are available to fetch from another device. Stubbed out some code for working on it, will investigate more first to see what the challenges are for that portion.

anatol: "fetch from another device" (either EFI or mountable USB) is a large task that will probably require some refactoring. I suggest leaving it for another pull request.

jacobmyers-codeninja:
Got pulled away on a business trip so didn't have a chance to finish cleaning this up. I agree that the actual read from external is a whole job on its own, I just wanted the placeholder for a future request to make easy integration work of it.

I have updated the docs and merged back in the new changes.

anatol:
@jacobmyers-codeninja I squashed your commits and cleaned up a bit. Please review it and then I'll move forward with merging it to the repo.

init/luks.go
```diff
@@ -298,6 +300,47 @@ func recoverTokenPassword(volumes chan *luks.Volume, d luks.Device, t luks.Token
...
+			warning("reading password: %v", err)
+		}
+	} else {
+		// read password from device matching uuid
+		uuid, err := parseUUID(parts[1])
+		if err != nil {
+			warning("invalid UUID %s in rd.luks.key boot param: %s", uuid, keyfile)
+		} else {
+			// TODO: access path on uuid device, read password
+			warning("user wants keyfile from device %s, but I don't know how do do that", uuid)
```
jacobmyers-codeninja, line 323: Looks like I made a typo here, "know how to do that" instead of "know how do do that"

init/main.go
```diff
@@ -36,7 +36,7 @@ var (
 
 	initBinary = "/sbin/init" // path to init binary inside the user's chroot
 
-	luksMappings []luksMapping // list of LUKS devices that booster unlocked during boot process
+	luksMappings []*luksMapping // list of LUKS devices that booster unlocked during boot process
```
jacobmyers-codeninja, line 39: I thought about this, but wasn't sure if it was a dangerous change or what else it might modify so that is why I went with the more sloppy solution. I like this much better though, thanks!

anatol:
Thank you for your work!

I landed the changes `wip` branch so I can test it at my machines before pushing it to master.

jacobmyers-codeninja:
> Thank you for your work!
> 
> I landed the changes `wip` branch so I can test it at my machines before pushing it to master.

Absolutely, let me know if you run into anything too. I will do my best to make sure it's good. I have a few machines here in different states I can test more with besides the one I was using to write it.

PR #167, opened 2022-08-14
Fix parsing errors when quotes are inside values

jacobmyers-codeninja:
For: https://github.com/anatol/booster/issues/73#issuecomment-1210899737

The parser is missing an edge case for quotes within values, which could be technically valid. This allows those conditions to pass successfully while keeping everything cleaned up as before. This is a very minimal modification to account for the condition. 

For now the error condition is reduced to allow quotes if they follow an equals sign. This allows the test case to work, but would still catch more unusual edge cases. If more conditions that need to allow quotes within sub-strings are needed the restriction on the quotes in values side can be removed to allow a better behavioral handle.

- Fixing parsing error that happens when quotes are deeper into values in key=value checks. IE: `key=UUID="test"`
- Update tests to account for coving the extra cases.
- Updated test cases to have better combinations to make looking for test failures easier
- Added an extra test for multi-parameter testing and validation, specifically putting in the failing luks test for now

### DIFF
--- a/init/cmdline_test.go
+++ b/init/cmdline_test.go
@@ -50,63 +50,100 @@ func TestGetNextParam(t *testing.T) {
 	}
 
 	tests := []test{
-		// \param00=test0 // an odd case, but we will allow it
-		{"\\param00=test0", "param00", "test0", 14},
-		// param01=test0
-		{"param01=test0", "param01", "test0", 13},
-		// "param02=test0"
-		{"\"param02=test0\"", "param02", "test0", 15},
-		// param03="test0"
-		{"param03=\"test0\"", "param03", "test0", 15},
-		// '   param04=test0   '
-		{"   param04=test0   ", "param04", "test0", 17},
-		// param05=te\0st
-		{"param05=te\000st0", "param05", "te", 11},
-		// [tab]param06=test0[tab]
-		{"\tparam06=test0\t", "param06", "test0", 15},
-		// param07=te"st0
-		{"param07=te\"st0", "param07", "te\"st0", 14},
-		// par"am08=test0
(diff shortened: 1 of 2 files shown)


init/cmdline_test.go
```diff
@@ -50,63 +50,100 @@ func TestGetNextParam(t *testing.T) {
...
+		{"param22=\"test22 param23=\"test23\"", "param22", "test22 param23=", 25},
+		// param24="test24 param25=test25 // This is a worst case bad junk input, it will mangle 24 and swallow 25
+		{"param24=\"test24 param25=test25", "param24", "test24 param25=test25", 30},
 	}
 
 	for _, test := range tests {
 		var k, v string
 		var i int
 
+		t.Log("Testing ", test.input, "\n")
```
anatol, line 107: no need to do the excessive printing

init/cmdline_test.go
```diff
@@ -50,63 +50,100 @@ func TestGetNextParam(t *testing.T) {
...
+				{"root", "UUID=7bbf9363-eb42-4476-8c1c-9f1f4d091385", 100},
+			},
+		},
+	}
+
+	for _, test := range tests {
+		var k, v string
+		i := 0
+
+		t.Log("Testing ", test.input)
```
anatol, line 141: ditto

init/cmdline_test.go
jacobmyers-codeninja: I believe the printing will only happen if the test fails, which makes tracking down the failure a lot easier. On success it shouldn't print anything unless run with verbose mode enabled.

I can remove it though if you would prefer.

anatol: Did not know about it.

Then the change looks good.

Issue #172, opened 2022-09-01
init tests fail because of new restriction on mkfs.xfs


thomasfsteeples:
Since 4 August 2022, `mkfs.xfs` has a new restriction, disallowing filesystems smaller than 300MB:

> commit 6e0ed3d19c54603f0f7d628ea04b550151d8a262
> Author: Darrick J. Wong <djwong@kernel.org>
> Date:   Thu Aug 4 21:27:01 2022 -0500
> 
>     mkfs: stop allowing tiny filesystems
>     
>     Refuse to format a filesystem that are "too small", because these
>     configurations are known to have performance and redundancy problems
>     that are not present on the volume sizes that XFS is best at handling.
>     
>     Specifically, this means that we won't allow logs smaller than 64MB, we
>     won't allow single-AG filesystems, and we won't allow volumes smaller
>     than 300MB.  There are two exceptions: the first is an undocumented CLI
>     option that can be used for crafting debug filesystems.
>     
>     The second exception is that if fstests is detected, because there are a
>     lot of fstests that use tiny filesystems to perform targeted regression
>     and functional testing in a controlled environment.  Fixing the ~40 or
>     so tests to run more slowly with larger filesystems isn't worth the risk
>     of breaking the tests.
>     
>     Signed-off-by: Darrick J. Wong <djwong@kernel.org>
>     Reviewed-by: Carlos Maiolino <cmaiolino@redhat.com>
>     Signed-off-by: Eric Sandeen <sandeen@sandeen.net> 

As such, `go test` fails for the init component of Booster. This is easily fixed by changing the test to create a filesystem of size 300MB rather than 100MB.

thomasfsteeples:
Fixed by #171.

PR #173, opened 2022-09-01
Split blkinfo test into multiple tests

thomasfsteeples:
When trying to diagnose https://github.com/anatol/booster/issues/172, I had (admittedly, not very much) trouble discerning which assertion was actually failing in `TestBlkInfo`. To this end, I propose splitting this test into multiple tests, allowing problems to be highlighted more easily, and adhering to the general maxim that only one logical component should be included per test.

### DIFF
--- a/init/blkinfo_test.go
+++ b/init/blkinfo_test.go
@@ -65,16 +65,39 @@ func fileExists(file string) bool {
 	return err == nil
 }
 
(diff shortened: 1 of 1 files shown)


anatol:
The change looks good. It has a conflict with the previous change though.

thomasfsteeples:
Cheers; have now merged earlier fix into this commit.

PR #174, opened 2022-09-01
Made bash scripts shellcheck-conformant, ran through formatter and tried to…

thomasfsteeples:
… remove a number of bash-isms. Mostly just quoting variable expansions, but a few other niceties are in there as well.

### DIFF
--- a/tests/generators/alpinelinux.sh
+++ b/tests/generators/alpinelinux.sh
@@ -1,11 +1,13 @@
+#!/usr/bin/env bash
+
 trap 'quit' EXIT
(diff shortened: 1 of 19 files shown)


anatol:
thanks, it looks good

actually could you please also provide what command did you use for formatting this files?

thomasfsteeples:
Sure, I used `shfmt -s -i 2 -sr`; `shfmt` is from the [sh project](https://github.com/mvdan/sh).

thomasfsteeples:
Thanks for merging this and the others!

PR #175, opened 2022-09-01
Moved all .gitignore content to root, and added binaries to it

thomasfsteeples:
Whilst git allows you to have multiple `.gitignore` files, the [recommendation/consensus online](https://stackoverflow.com/questions/10274424/best-practice-for-using-multiple-gitignore-files) seems to be that it's better just to stick with one root `.gitignore`. I've moved the contents of all the `.gitignore`s to the root `.gitignore`, and also added the `init` and `generator` binaries to it (which are currently sat in the repo).


anatol:
oof, binaries in the repo is not good. Thanks for fixing it.

PR #178, opened 2022-09-03
Deduplicated check in TestExtraFiles

thomasfsteeples:
In `TestExtraFiles` in `generator_test.go`, a `stat` check was being performed twice --- once inside the `for` loop and then again afterwards with calls to `checkFileExistence`. Have deleted the latter checks and converted the check inside the `for` loop to a call to `checkFIleExistence`. Also converted string concatenation of paths to calls to `filepath.Join`.


anatol:
Thank you for this PR!

PR #179, opened 2022-09-03
Tweaked symlinks so unpacked paths point to correct locations

thomasfsteeples:
Currently, in the `generator` test suite, I have several tests failing, including ` TestExtraFiles`, ` TestEnableVirtualConsole`, `TestEnableVirtualConsoleWithoutLocaleConf`. This seems to be because the symlinks are currently set up to point to the root directory. Whilst this is fine when used in practice as an initramfs, it means that when unpacked, it points to the root directory of the host machine (which is not the intended effect). Removing the leading slash makes the test suite pass, and the resulting image remains bootable.

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -243,16 +243,16 @@ func generateInitRamfs(conf *generatorConfig) error {
 // appendCompatibilitySymlinks appends symlinks for compatibility with older firmware that loads extra files from non-standard locations
 func appendCompatibilitySymlinks(img *Image) error {
 	symlinks := []struct{ src, target string }{
-		{"/lib", "/usr/lib"},
-		{"/usr/local/lib", "/usr/lib"},
-		{"/usr/sbin", "/usr/bin"},
-		{"/bin", "/usr/bin"},
-		{"/sbin", "/usr/bin"},
-		{"/usr/local/bin", "/usr/bin"},
-		{"/usr/local/sbin", "/usr/bin"},
-		{"/var/run", "/run"},
-		{"/usr/lib64", "/usr/lib"},
-		{"/lib64", "/usr/lib"},
+		{"/lib", "usr/lib"},
+		{"/usr/local/lib", "usr/lib"},
+		{"/usr/sbin", "usr/bin"},
+		{"/bin", "usr/bin"},
+		{"/sbin", "usr/bin"},
+		{"/usr/local/bin", "usr/bin"},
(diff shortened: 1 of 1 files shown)


generator/generator.go
```diff
@@ -243,16 +243,16 @@ func generateInitRamfs(conf *generatorConfig) error {
 // appendCompatibilitySymlinks appends symlinks for compatibility with older firmware that loads extra files from non-standard locations
 func appendCompatibilitySymlinks(img *Image) error {
 	symlinks := []struct{ src, target string }{
```
anatol, line 245: So you are basically using relative symlinks. It is better to change the commit description "Tweaked symlinks" -> "Use relative symlinks".

generator/generator.go
```diff
@@ -243,16 +243,16 @@ func generateInitRamfs(conf *generatorConfig) error {
...
-		{"/lib", "/usr/lib"},
-		{"/usr/local/lib", "/usr/lib"},
```
anatol, line 247: the target needs to be "../usr/lib"

generator/generator.go
```diff
@@ -243,16 +243,16 @@ func generateInitRamfs(conf *generatorConfig) error {
...
-		{"/usr/sbin", "/usr/bin"},
```
anatol, line 248: target needs to be "../usr/bin"

generator/generator.go
```diff
@@ -243,16 +243,16 @@ func generateInitRamfs(conf *generatorConfig) error {
...
-		{"/bin", "/usr/bin"},
-		{"/sbin", "/usr/bin"},
-		{"/usr/local/bin", "/usr/bin"},
```
anatol, line 251: ditto

generator/generator.go
```diff
@@ -243,16 +243,16 @@ func generateInitRamfs(conf *generatorConfig) error {
...
-		{"/usr/local/sbin", "/usr/bin"},
```
anatol, line 252: ditto.

here and any other place where symlinks cross directory hierarchy

anatol (changes requested):

anatol:
@thomasfsteeples I would like to take care of your changes and move forward. Are you OK with it?

thomasfsteeples:
@anatol Absolutely fine with me; sorry I never got round to it!

anatol:
Pushed this PR to `wip` branch as 209f8cce95ab08c9c5f840aaff0a7ab61f1ac89c

PR #180, opened 2022-09-04
Changed kernel modules/firmware directories from /usr/lib/ to /lib/; also tidied Arch packaging scripts in the process

thomasfsteeples:
In the Linux kernel, the default modules directory is `/lib/modules` rather than `/usr/lib/modules`. The reason the current code works on many distros, such as Arch Linux, Fedora and Debian, is because these distribution have merged the members of the `/usr` directory with their root equivalents ([Arch merge](https://archlinux.org/news/the-lib-directory-becomes-a-symlink/); [Fedora merge](https://github.com/anatol/booster/pull/url); [Debian merge](https://wiki.debian.org/UsrMerge)). However, not all distros have performed this merge (like Gentoo), and the [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html) still regards them as separate.

This proposed commit will preserve current behaviours on systems with a usrmerge hierarchy (as the symlink will just be followed); will now act correctly on systems without a usrmerge and where usr does not sit on a separate partition; and will open up the possibility of extending support to those systems without a usrmerge and with separate usr partitions. All tests remain passing on my Arch Linux system, and have gone from none of them passing to (almost) all of them passing on my Gentoo system.

Further reading if interested: [1](https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/), [2](https://freedesktop.org/wiki/Software/systemd/separate-usr-is-broken/), [3](https://lwn.net/Articles/890219/).

### DIFF
--- a/generator/generator_test.go
+++ b/generator/generator_test.go
@@ -324,9 +324,9 @@ func TestUniversalMode(t *testing.T) {
 	require.Equal(t, "matestkernel", cfg.Kernel)
 
 	// all except kernel/testfoo.ko need to be in the image
-	checkDirListing(t, opts.workDir+"/image.unpacked/usr/lib/modules/", "foo.ko", "cbc.ko", "virtio_scsi.ko", "booster.alias")
+	checkDirListing(t, filepath.Join(opts.workDir, "/image.unpacked/lib/modules/"), "foo.ko", "cbc.ko", "virtio_scsi.ko", "booster.alias")
 
-	aliasesFile, err := os.ReadFile(opts.workDir + "/image.unpacked/usr/lib/modules/booster.alias")
+	aliasesFile, err := os.ReadFile(filepath.Join(opts.workDir, "/image.unpacked/lib/modules/booster.alias"))
 	require.NoError(t, err)
 
 	expectedAliases := `pci:v*d*sv*sd*bc0Csc03i30* cbc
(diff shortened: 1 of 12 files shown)


generator/generator_test.go
```diff
@@ -335,9 +335,9 @@ cpu:type:x86,ven*fam*mod*:feature:*0081* cbc
 `
 	require.Equal(t, expectedAliases, string(aliasesFile))
 
-	checkFileExistence(t, opts.workDir+"/image.unpacked/usr/lib/firmware/whiteheat.fw.xz")
-	checkFileExistence(t, opts.workDir+"/image.unpacked/usr/lib/firmware/usbdux_firmware.bin.xz")
-	checkFileExistence(t, opts.workDir+"/image.unpacked/usr/lib/firmware/rtw88/rtw8723d_fw.bin.xz")
+	checkFileExistence(t, filepath.Join(opts.workDir, "/image.unpacked/lib/firmware/whiteheat.fw.xz"))
```
anatol, line 338: filepath.Join() should probably be part of another change. Probably in the same commit as 'Converted string concanetation to call to filepath.Join in TestExtraFiles'

packaging/arch/booster-install
```diff
@@ -4,30 +4,30 @@ kernels=()
 all=0
 
 while read -r line; do
-    # line is like usr/lib/modules/5.8.2-arch1-1/vmlinuz
```
anatol, line 7: bash reformatting this should be in a separate commit as well

anatol (changes requested):
You have a lot of good information in the PR description. Links to the "/usr" merge documents in particular. Could you please move it to the commit description to make it easier to understand the change only using `git log`.?

I do not mind to move to non-/usr config if it helps the users. 

Also note that distros have different paths conventions and configure it using the `ldflags`

Here is the Linux Apline changes https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/booster/APKBUILD
```
--ldflags="-X main.imageModulesDir=/lib/modules/ \
		-X main.firmwareDir=/lib/firmware/ \
		-X main.consolefontsDir=/usr/share/consolefonts/"
```

anatol:
hi @thomasfsteeples I would like to move forward with these changes. Do you still work on it?

thomasfsteeples:
Hi @anatol --- sorry, been busy with work lately, but I'll try and have a crack with splitting into separate commits in the next few days, and then I'll resubmit. Does that work for you?

anatol:
@thomasfsteeples yes it works for me

anatol:
also cc @nmeum

anatol:
Pushed to `wip` branch as b040e83b9bf4c28082589ccee538cbec3cfe82a3

anatol:
It turns out this PR beaks integration tests. CPIO compressor does not like to have 'lib as a symlink' together with `/lib/modules/...` files.

`Image.Append*()` needs to perform path sanitization and handle symlinks.

Or another way to fix the issue is to modify `generator/image.go`. Instead of implementing own streaming cpio/zstd compressor it is possible handle in a more traditional way - write all the files to a temp dir first and then use command line tools to do cpio/zstd compression. This way the code avoids dealing path sanitizing.

PR #187, opened 2022-09-26
init: allow force loading of extra modules via the kernel command-line

### LINKED ISSUE #186: Provide a way to force load modules with having to rebuild the image
While debugging #185 I noticed that there doesn't seem to be any way to force load a module without rebuilding the image. As such, even though I knew which modules I had to load in order to workaround #185 I still had to boot from a USB stick, modify `modules_force_load` in `/etc/booster.yaml`, rebuild the initramfs and reboot. I think in such situation it would be convenient to provide a way to force load modules before the init is run.

One easy way to achieve that would via a Kernel parameter, e.g. `modules=t10_pi,sd_mod`. If this would be supported, I would just need to edit the Kernel parameters from my bootloader configuration menu. Alternatively, it might be useful to provide a way to re-run the Booster mounting code after the BusyBox emergency shell has exited. This would allow users to manually load modules using `busybox insmod`.

Thoughts?

nmeum:
Fixes #186

### DIFF
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -129,6 +129,7 @@ Some parts of booster boot functionality can be modified with kernel boot parame
    If debug level is enabled then kmsg throttling gets disabled automatically.
  * `booster.debug` an obsolete option that is equivalent to `booster.log=debug,console`.
  * `quiet` Set booster init verbosity to minimum. This option is ignored if `booster.debug` or `booster.log` is set.
+ * `modules` a comma-separated list of extra kernel modules which should be force loaded.
  * `init=$PATH` path to user-space init binary. If not specified then default value `/sbin/init` is used.
 
 ## NOTES
--- a/init/cmdline.go
+++ b/init/cmdline.go
@@ -192,7 +192,13 @@ func parseParams(params string) error {
 			if err != nil {
 				return fmt.Errorf("root=%s: %v", value, err)
 			}
+		case "modules":
+			if (value == "") {
+				break
+			}
 
+			modules := strings.Split(value, ",")
+			config.ModulesForceLoad = append(config.ModulesForceLoad, modules...)
 		case "resume":
 			var err error
 			cmdResume, err = parseDeviceRef(value)


init/cmdline.go
```diff
@@ -192,7 +192,14 @@ func parseParams(params string) error {
 			if err != nil {
 				return fmt.Errorf("root=%s: %v", value, err)
 			}
+		case "modules":
+			if (value == "") {
+				break
+			}
 
+			for _, o := range strings.Split(value, ",") {
+				config.ModulesForceLoad = append(config.ModulesForceLoad, o)
```
anatol, line 201: I think shorter version should work here as well

```go
modules := range strings.Split(value, ",")
config.ModulesForceLoad = append(config.ModulesForceLoad, modules...)
```

docs/manpage.md
```diff
@@ -129,6 +129,7 @@ Some parts of booster boot functionality can be modified with kernel boot parame
    If debug level is enabled then kmsg throttling gets disabled automatically.
  * `booster.debug` an obsolete option that is equivalent to `booster.log=debug,console`.
  * `quiet` Set booster init verbosity to minimum. This option is ignored if `booster.debug` or `booster.log` is set.
+ * `modules` a comma-separated list of extra kernel modules which should be force loaded.
```
anatol, line 132: what parameter name is used by other initramfs? I vaguely remember dracut likes to use `rd.` prefix for the parameters used by initramfs only.

anatol (commented):
The feature looks good. Just a few questions to clarify.

Later I will add an integration test to check that the boot param works as intended.

docs/manpage.md
nmeum: * Alpine's mkinitfs uses `modules`, which is why I ended up using it here as well but I also wouldn't mind using something else of cause
* dracut seems to distinguish `rd.driver.pre` and `rd.driver.post`: https://github.com/dracutdevs/dracut/blob/master/man/dracut.cmdline.7.asc#misc
* mkinitcpio seems to use `earlymodules`: https://wiki.archlinux.org/title/Mkinitcpio#init_from_base_hook

init/cmdline.go
nmeum: Thanks, good idea. Adjusted accordingly. Haven't tested it with this change but I assume it should work. On a side note: the `if value == ""` is necessary as otherwise Booster would attempt to load `/usr/lib/modules/.ko` with a `modules=` kernel parameter.

docs/manpage.md
anatol: I propose to use the dracut's idea with `rd.` prefix for initrd-specific parameters. I am going to merge your change to `wip` and modify the parameter name.

anatol: Or maybe even call it `rd. modules_force_load` to match the name to config option that does the same.

PR #197, opened 2022-11-29
generator: normalize both module and pattern names from modules.alias

### LINKED ISSUE #185: Booster fails to load sd_mod.ko
After a recent system upgrade, Booster failed to prompt for my FDE password. I investigated this further and noticed that the device files for my SATA devices are not present in the BusyBox emergency environment. Hence, Booster does not find my encrypted LUKS partition and thus doesn't prompt for a password. The device drivers for my SATA device require the `sd_mod` kernel module to be loaded by Booster, which it doesn't seem to do automatically anymore for some reason.

Manually loading `sd_mod` in the BusyBox emergency shell results in the following error:

```
# busybox insmod /usr/lib/modules/sd_mod.ko
sd_mod: Unknown symbol t10_pi_type1_crc (err -2)
sd_mod: Unknown symbol t10_pi_type3_crc (err -2)
sd_mod: Unknown symbol t10_pi_type3_ip (err -2)
…
```

As it turns out, these symbols are provided by `t10_pi.ko` and indeed loading `t10_pi.ko` first and then loading `sd_mod.ko` works as expected:

```
# busybox insmod /usr/lib/modules/t10_pi.ko
# busybox insmod /usr/lib/modules/sd_mod.ko
< sd* devices appear in /dev >
```

According to `modinfo`, the `sd_mod` module also has a direct dependency on `t10_pi.ko` on my system:

```
$ modinfo sd_mod | grep ^depends
depends:        scsi_mod,t10-pi
```

As such, maybe this issue is similar to #133 in the sense that dependencies are not correctly taken into account and therefore `sd_mod.ko` is not loaded by Booster and hence it doesn't prompt for the LUKS password? This hypothesis seems to be confirmed by the fact that using the following `modules_force_load` declaration in `/etc/booster.yaml` makes Booster work again:

```
modules_force_load: t10_pi,sd_mod
```

As I mentioned above, I encountered this after a system upgrade. The system booted fine with Booster in the past without the `modules_force_load` declaration. However, I am not sure if this is a regression introduced in Booster 0.9 or if this is due to a recent Kernel upgrade.

nmeum:
From `modprobe.d(5)`:

> Note that module and alias names (like other module names) can have - or _ in them: both are interchangeable throughout all the module commands as underscore conversion happens automatically.

For this reason, we need to make sure that everywhere where module or pattern names are compared that this comparison is always performed on a normalized string. This commit adds several normalizations for module and pattern names that were previously not performed.

Fixes #185

### DIFF
--- a/generator/kmod.go
+++ b/generator/kmod.go
@@ -250,6 +250,12 @@ func (k *Kmod) readKernelAliases() error {
 		}
 		pattern := line[:idx]
 		module := line[idx+1:]
+
+		// As per modprobe.d(5), - and _ can be used
+		// interchangeably in both module and alias names.
+		pattern = normalizeModuleName(pattern)
+		module = normalizeModuleName(module)
+
 		k.aliases = append(k.aliases, alias{pattern, module})
 	}
 
@@ -704,7 +710,7 @@ func readDeviceAliases() (set, error) {
 		if err != nil {
 			return err
 		}
-		al := strings.TrimSpace(string(b))
+		al := normalizeModuleName(strings.TrimSpace(string(b)))
 		if al == "" {
 			return nil
 		}
--- a/init/udev.go
+++ b/init/udev.go
@@ -109,7 +109,7 @@ func handleUdevEvent(ev netlink.UEvent) {
 	debug("udev event %+v", ev)
 
 	if modalias, ok := ev.Env["MODALIAS"]; ok {
-		go func() { check(loadModalias(modalias)) }()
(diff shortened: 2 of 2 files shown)


anatol:
Thank you for the fix. This change looks good. Merging it to `wip`, will test it at my machines for a couple of days, and then land it `master`.

anatol:
All tests look good. I merged this fix to `master`.

Thank you for the fix!

PR #209, opened 2023-03-09
unpack: Improve symlink handling

nmeum:
This commit fixes `booster unpack` issues which I noticed after upgrading my Alpine Linux booster version to 0.10. This release changes directory symlinks (e13aa77e49aef93cc1370a269785b471758cb546) to relative paths. As such, we need to make sure that we change into the output directory before attempting to resolve them. Furthermore, we need to account for symlinks during unpack, i.e. when unpacking `bin/busybox` (where `bin/` is a relative symlink to `usr/bin`) we need to resolve this to `usr/bin/busybox` first. For this purpose, we can't use `filepath.EvalSymlinks` as it does not handle broken symlinks well (e.g. if `bin/busybox` is unpacked before `usr/bin` is created).

### DIFF
--- a/generator/unpack.go
+++ b/generator/unpack.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"compress/gzip"
+	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


generator/unpack.go
```diff
@@ -88,9 +90,34 @@ func processImage(file string, fn processCpioEntryFn) error {
...
+
 	fn := func(hdr *cpio.Header, r *cpio.Reader) error {
 		out := filepath.Join(dir, hdr.Name)
-		if err := os.MkdirAll(filepath.Dir(out), 0o755); err != nil {
+		dirFp := filepath.Dir(out)
+
+		// Resolve symlinks for the dirFp and update out
+		// accordingly (e.g. bin/busybox → usr/bin/busybox).
+		//
+		// XXX: Can't use filepath.EvalSymlinks as it emits
```
anatol, line 107: What is `XXX:` is it a TODO?

anatol (approved):
Looks good. I will test locally real quick.

nmeum:
Just noticed that there is another bug with https://github.com/anatol/booster/commit/e13aa77e49aef93cc1370a269785b471758cb546: Some target directories (e.g. `usr/lib` are not created if the host system doesn't ship any files in them). This explains the broken symlinks I saw during unpack. Maybe the handling for that is not even need with #209 applied?

nmeum:
> Maybe the handling for that is not even need with #210 applied?

Yep, it seems that with #210 this is no longer needed as the broken symlinks described here only existed because some directories (e.g. `usr/lib`) did not exist in the image generated by Booster. This should also explain why `booster unpack` still passed the tests on Arch while it did not work on Alpine.

anatol:
Thank you for the clarification. I will look at #210

PR #211, opened 2023-03-10
booster cat: return an error if the requested file doesn't exist

nmeum:
Presently, it is not possible to distinguish between an empty file and a non-existent file based on the `booster cat` output and exit status. With this patch applied, booster prints an error message to standard error and exists with a non-zero exit status if the file was not found in the image.

### DIFF
--- a/generator/unpack.go
+++ b/generator/unpack.go
@@ -4,6 +4,7 @@ import (
 	"compress/gzip"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


anatol:
Thank you for the patch! It looks good to me.

I should look at refactoring/simplifying the unpack.go code and use a bit more straightforward (no function pointers) and simple logic there.

PR #214, opened 2023-03-23
Resolve symlinks when adding extra files

### LINKED ISSUE #213: initramfs fails to build image with `lvm` in $PATH
When booting with this initramfs, it fails because it can't find the `lvm` executable in it's `$PATH`. Extracting the generated initramfs using `unzstd`, followed by `cpio -i`, shows that booster did not include the `lvm` binary in the initramfs, despite there being an entry for it in the cpio archive
```
cpio: bin/lvm: Cannot open: No such file or directory
```
This is my config
```yaml
modules_force_load: amdgpu
universal: false
compression: zstd
strip: true
modules: kernel/arch/x86/crypto/,kernel/drivers/crypto/,kernel/crypto/
enable_lvm: true
busybox: true
```
Adding the config value
```yaml
extra_files: /usr/bin/lvm
```
Seems to fix the issue.

This config value
```yaml
extra_files: lvm
```
Does not fix the issue

Also worth noting is that booster fails to create a /usr/bin directory as well, which results in a broken `/bin` symlink. This problem may be the result of booster's generator trying to to put the `lvm` binary at `bin`, which is a broken symlink

StratusFearMe21:
This should fix #213 (fixes it for me). I don't know go, so I'm sorry if my code is a little messy.

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -289,10 +289,15 @@ func (img *Image) appendExtraFiles(binaries ...string) error {
 		if !filepath.IsAbs(f) {
 			// If the given name is not an absolute path, assume that it refers
 			// to an executable and lookup the path to the executable using $PATH.
-			var err error
-			f, err = exec.LookPath(f)
-			if err != nil {
-				return err
+			var err_lp error
+			var err_abs error
+			f, err_lp = exec.LookPath(f)
+			f, err_abs = filepath.Abs(f)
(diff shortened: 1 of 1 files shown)


StratusFearMe21:
@anatol does 7b4011a3cae1c337f3aeefba0596beef7565a0a7 overrule this PR?

Issue #224, opened 2023-07-02
Handle zstd compressed firmware


evelikov:
Coming from https://bugs.archlinux.org/task/78892:

In seems like booster does not know about zstd compressed firmware, even though the feature landed in upstream kernel with 5.19 - about 11 months ago. Which in turn means that modules will barf up, since their firmware isn't present in the initrd.

At a glance it seems to be a one-liner - will open a MR in a moment.

PR #225, opened 2023-07-02
Handle zstd compressed firmware

evelikov:
Nothing special here - just another extension to handle. While we're here, mentioning the kernel version which landed support is more beneficial to when it was by distribution X or Y - the project is aiming at multiple distributions after all.

Closes: https://github.com/anatol/booster/issues/224

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -331,6 +331,7 @@ func findFwFile(fw string) (string, error) {
 	supportedFwExt := []string{
 		"",
 		".xz", // since linux v5.3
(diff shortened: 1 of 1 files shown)


anatol:
Looks great, thank you!

evelikov:
Thank you for the quick response.

IMHO it would be nice to have tests handling all the different extensions. At a glance it seems like everything is hard-coded to XZ. Something to perhaps look into as follow-up and nerd-snipe people ... not it ;-)

PR #239, opened 2023-08-13
packaging: arch: Do not report removal errors for nonexistent files

Zile995:
Some users may delete booster images if they create UKI files.
Do not report removal errors if file does not exist.

mkinitcpio deletes files with the -f option:
https://github.com/archlinux/mkinitcpio/blob/ffae4b4821aa7d12271165cacf87d6772c70a8b4/libalpm/scripts/mkinitcpio#L115


anatol:
Thank you for the contribution

PR #245, opened 2023-09-01
Add debian lib64 path to elfLibDir

ferrreo:
Hi,

I've recently been trying to packaging booster for my debian/ubuntu based distro and we ran into an issue with the elfLibDir not having the path debian uses. This PR fixes that.


anatol:
Thank you!

Issue #246, opened 2023-09-28
systemd-boot now defaults `/efi` as the folder for initramfs images instead of `/boot`. How does booster cope with that?


enihcam:
I'm asking this question because I see some hardcoded `/boot` in booster codebase.

enihcam:
As you can see, [dracut](https://github.com/dracutdevs/dracut) fixed this already, so maybe you can take the following code as a reference?
https://github.com/dracutdevs/dracut/commit/093bc9b5cfc071409a20351a669c594c748eb647
https://github.com/dracutdevs/dracut/commit/49ea6c42db7180eec5ba57e082a38d116f2d17a5

enihcam:
Instead of ad-hoc script, I thought an elegant way to copy initramfs artifacts over to $BOOT directory is to use [systemd-kernel-install](https://www.freedesktop.org/software/systemd/man/kernel-install.html)

An example would be:
https://gitlab.com/dalto.8/kernel-install-for-dracut

What do you think?

c3Ls1US:
On Arch Linux, Booster provides a shell script that puts the initramfs image at /boot. The shell script does not currently accommodate this causing the image to not be built. As a result, since there's no image, the bootloader probably wouldn't find it.

As a workaround, you would have to manually build and specify the appropriate path

enihcam:
> On Arch Linux, Booster provides a shell script that puts the initramfs image at /boot. The shell script does not currently accommodate this which (should) cause the image to not be built. As a result, since there's no image, this implies that the bootloader wouldn't find it.
> 
> As a workaround, you would have to manually build and specify the appropriate path

right, that's the problem. as a workaround, i manually replaced `/boot` with `/efi`. hopefully there'll be a more elegant approach to resolve this.

c3Ls1US:
@enihcam To follow up, is there any specific reason why you do not have a `/boot` directory? I understand the case where one can have both `/boot` and `/efi` but not only `/efi`.

This is an Arch Linux script so the default boot path for initramfs images and other components is `/boot`.  If going by mkinitcpio and assuming you have only 1 ESP partition, `/efi` seems to be only reserved for UKIs which (should) require the user to adjust the new mount point of the ESP to `/efi`, leaving boot files residing within the root partition.

[I understand that there are cases for other ESP mount points](https://wiki.archlinux.org/title/EFI_system_partition#Typical_mount_points), but what's yours specifically?

kjkent:
[...]

PR #247, opened 2023-09-29
arch: handle more boot paths for installs

c3Ls1US:
Fixes https://github.com/anatol/booster/issues/246

### DIFF
--- a/packaging/arch/regenerate_images
+++ b/packaging/arch/regenerate_images
@@ -3,6 +3,9 @@
 # find out all installed kernels
 mapfile -d '' kernels < <(find /usr/lib/modules -maxdepth 1 -type d ! -name "modules" -print0)
 
+# paths to boot partition for the initramfs image and kernel
+boot_paths=("/efi" "/boot")
+
 for kernel in "${kernels[@]}"; do
   if ! pacman -Qqo "${kernel}/pkgbase" > /dev/null 2>&1; then
     # if pkgbase does not belong to any package then skip this kernel
@@ -10,8 +13,15 @@ for kernel in "${kernels[@]}"; do
   fi
   read -r pkgbase < "${kernel}/pkgbase"
 
-  booster build --force --kernel-version ${kernel##/usr/lib/modules/} /boot/booster-${pkgbase}.img &
-  install -Dm644 "${kernel}/vmlinuz" "/boot/vmlinuz-${pkgbase}"
+  # build and install the image and kernel at one boot path
+  for path in "${boot_paths[@]}"; do
+    if [ -d "${path}" ] && [ -w "${path}" ]; then
+      booster build --force --kernel-version "${kernel##/usr/lib/modules/}" "$path"/booster-"${pkgbase}".img &
+      install -Dm644 "${kernel}/vmlinuz" "${path}/vmlinuz-${pkgbase}"
+
+      break
+    fi
+  done
 done
 
 wait
--- a/packaging/arch/regenerate_images
+++ b/packaging/arch/regenerate_images
@@ -13,13 +13,11 @@ for kernel in "${kernels[@]}"; do
   fi
   read -r pkgbase < "${kernel}/pkgbase"
 
-  # build and install the image and kernel at one boot path
+  # build and install the image and kernel at all boot paths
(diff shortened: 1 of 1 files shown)


packaging/arch/regenerate_images
```diff
@@ -3,15 +3,25 @@
...
 
-  booster build --force --kernel-version ${kernel##/usr/lib/modules/} /boot/booster-${pkgbase}.img &
-  install -Dm644 "${kernel}/vmlinuz" "/boot/vmlinuz-${pkgbase}"
+  # build and install the image and kernel at one boot path
+  for path in "${boot_paths[@]}"; do
+    if [ -d "${path}" ] && [ -w "${path}" ]; then
+      booster build --force --kernel-version "${kernel##/usr/lib/modules/}" "$path"/booster-"${pkgbase}".img &
+      install -Dm644 "${kernel}/vmlinuz" "${path}/vmlinuz-${pkgbase}"
+
+      break
```
c3Ls1US, line 22: Assumes that the image and kernel are in the same directory and I don't know if that's always the case?

packaging/arch/regenerate_images
```diff
@@ -3,15 +3,25 @@
 # find out all installed kernels
 mapfile -d '' kernels < <(find /usr/lib/modules -maxdepth 1 -type d ! -name "modules" -print0)
 
+# paths to boot partition for the initramfs image and kernel
+boot_paths=("/efi" "/boot")
```
c3Ls1US, line 7: `/boot/efi` is technically also a valid boot partition path for systemd and from the mkinitcpio presets it appears Arch Linux assumes `/efi` by default. Since paths should be distribution dependent, choosing not to include `/boot/efi`

anatol (approved):

anatol:
Thank you! The change looks reasonable.

packaging/arch/regenerate_images
c3Ls1US, line 22: I believe this is wrong and we shouldn't `break`. In the case that `/efi` exists and is empty, and the boot path of the user is **actually** `/boot`, then neither the image and the kernel will not be installed.

Besides being lazy and installing them at all paths, is there any way to determine the users' boot path?

Zile995: You can get the ESP or BOOT paths with the `bootctl` command, something like this:
```Shell
esp_path="$(bootctl -p)"
boot_path="$(bootctl -x)"
```

If the `XBOOTLDR` partition is used, esp_path and boot_path should be different

Zile995:
The Arch Linux user actually needs to choose whether `ESP` will be mounted to `/efi` or `/boot`, ie. to decide whether initramfs will be installed on the `ESP` or the `/boot` directory (root partition)

For example, `amd-ucode` and `intel-ucode` are always installed in the `/boot` directory...
https://gitlab.archlinux.org/archlinux/packaging/packages/linux-firmware/-/blob/main/PKGBUILD?ref_type=heads#L111
https://gitlab.archlinux.org/archlinux/packaging/packages/intel-ucode/-/blob/main/PKGBUILD?ref_type=heads#L28
`mkinitcpio` also installs initramfs there by default, right?

So why change?

packaging/arch/regenerate_images
c3Ls1US: Thanks but that won't work in all of the cases. `bootctl` is dependent and part of the systemd project.

c3Ls1US:
> The Arch Linux user actually needs to choose whether `ESP` will be mounted to `/boot` or `/efi` , ie. to decide whether initramfs will be installed on the `ESP` or the `/boot` directory (root partition)
> 
> For example, `amd-ucode` and `intel-ucode` are always installed in the `/boot` directory... https://gitlab.archlinux.org/archlinux/packaging/packages/linux-firmware/-/blob/main/PKGBUILD?ref_type=heads#L111 https://gitlab.archlinux.org/archlinux/packaging/packages/intel-ucode/-/blob/main/PKGBUILD?ref_type=heads#L28 `mkinitcpio` also installs initramfs there by default, right?
> 
> So why change?

I'm closing this PR in favor of aligning to the distribution's current default paths. Going against the distro's standards just makes life harder and not having `/boot` doesn't appear to be common.

Issue #110, opened 2021-10-12
Add support for secure boot


anatol:
Add a way to generate booster-based unified images that get signed with a key. Such final images could be used for machines configured with secure boot.

For the reference here is secure boot implementation from mkinitcpio project:
https://github.com/archlinux/mkinitcpio/commit/f40bb42577284083adaf1edca2f05300badb78b2

j0057:
It turned out to be pretty easy to do this with sbupdate; it takes care of generating the EFI Unified Kernel Image, with firmware included, and then signing it using `sbsign`.

This sbupdate.conf takes initrd images in /boot made with mkinitcpio as well as booster for both linux and linux-zen kernels, and puts everything in /efi:

```
KEY_DIR=/etc/efi-keys
ESP_DIR=/efi
OUT_DIR=
SPLASH=/dev/null
BACKUP=5
CMDLINE_DEFAULT='root=/dev/mapper/crypt_root resume=/dev/mapper/crypt_swap rw'

CONFIGS=(['linux']='archlinux archlinux-fallback archlinux-booster'
         ['linux-zen']='archlinux-zen archlinux-zen-fallback archlinux-zen-booster')

INITRD=(['archlinux']='/boot/initramfs-linux.img'
        ['archlinux-fallback']='/boot/initramfs-linux-fallback.img'
        ['archlinux-booster']='/boot/booster-linux.img'
        ['archlinux-zen']='/boot/initramfs-linux-zen.img'
        ['archlinux-zen-fallback']='/boot/initramfs-linux-zen-fallback.img'
        ['archlinux-zen-booster']='/boot/booster-linux-zen.img')

CMDLINE=(['archlinux-booster']='root=/dev/disk/by-label/luks_root resume=/dev/disk/by-label/luks_swap rw'
         ['archlinux-zen-booster']='root=/dev/disk/by-label/luks_root resume=/dev/disk/by-label/luks_swap rw')

EXTRA_SIGN=('/efi/shellx64-signed.efi')
```

anatol:
These are great news @j0057. I like the way this EFI generation decoupled from initramfs generation. Does `sbupdate` hook take care of ordering i.e. it waits for booster-initrmafs.img generated first and only then EFI binary is built/signed?

Are there any `sbupdate` equivalents at Arch community repo? If not should `sbupdate` be moved to `[community]` so people would be able to use it with less efforts?

j0057:
[...]

Issue #79, opened 2021-04-12
Support UEFI Executable generation with systemd stubs


Foxboron:
This is probably a nice weekend hack for someone, but generating UEFI executables is a must have feature for initramfs generators these days. My WIP patch for `mkinitcpio` serves as a very nice outline.

Someone that want to try contribute to this project can probably take a stab at it :) Else I can implement it once I have less on my todo list. I'll help review it if needed!

https://github.com/archlinux/mkinitcpio/pull/53

anatol:
Yes the idea is interesting.

Looking at your patch - how do you sign the result efi boot image?

Foxboron:
dracut allows you to specify a signing key and it uses `sbsign`. For `mkinitcpio` I'd rather implement some "post image generation" hook that would let you use [sbctl](https://github.com/Foxboron/sbctl/), but personally this is run with a separate pacman hook for `sbctl` currently.

Foxboron:
If you do want to use golang for this, you could consider using `go-uefi`.

https://github.com/Foxboron/go-uefi/blob/master/cmd/gosign/main.go

But the API isn't really stable and needs a lot more work.

anatol:
Yeah, a golang lib is preferable. API stability is not much of concern (thanks to golang modules).

c3Ls1US:
Same topic as https://github.com/anatol/booster/issues/110. It doesn't necessarily need to be in Booster.

thePrivacyFanatic:
since the code is distro specific maybe [booster uki manager](https://github.com/Zile995/booster-um) would solve it better?

PR #252, opened 2024-01-22
arch: add script to generate UKIs

c3Ls1US:
Currently, booster does not support generating [Universal Kernel Images](https://uapi-group.org/specifications/specs/unified_kernel_image/) (UKIs) natively and this PR allows creating UKIs via ukify on Arch Linux. 

To generate UKIs, we can simply delegate the functionality to external tools such as ukify. The resulting binaries are currently stored in `/boot/Linux/EFI` and manual intervention would be required from the user to configure the appropriate boot loader configution files to load the binary by default. Otherwise, binaries would probably need to be selected at boot time. For instance, in systemd-boot, `loader.conf` should be modified.

### DIFF
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -250,6 +250,10 @@ So /boot/loader/entries/booster.conf should looks like this:
     initrd /booster-linux.img
     options root=UUID=69bc4dd2-7f6c-4821-aa6b-d80d9c97d470 rw rootflags=relatime,autodefrag,compress=zstd:2,space_cache,subvol=root
 
+Create a Unified Kernel Image and write the result to /boot/EFI/Linux:
+
+    $ /usr/lib/booster/regenerate_uki build /boot/EFI/Linux
+
 ## COPYRIGHT
 Booster is Copyright (C) 2020 Anatol Pomazau <http://github.com/anatol>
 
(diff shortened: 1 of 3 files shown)


anatol:
Thank you! I am sure a lot of people find UKI functionality useful.

Could you please add a section to the documentation (https://github.com/anatol/booster/blob/master/docs/manpage.md) about what is UKI and how to use booster with it?

anatol:
Looks great, thank you!

c3Ls1US:
No problem. Also, one thing to consider is that the result is written to `/boot/Linux/EFI` by default.

Some users have a separate partition (e.g. `/efi`) where they boot EFI executables, which is a problem. Arch Linux also assumes `/efi` [by default in its own initrd generator](https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/blob/master/mkinitcpio.d/example.preset). For flexibility, I'm considering of changing it so a boot path is required to pass to the script.

anatol:
Sounds good.

Let me know once you are done developing the PR and I move with merging it.

c3Ls1US:
@anatol let me know if you prefer squashing my commits.

anatol:
I will squash the commits before merging

c3Ls1US:
Finished

This PR closes https://github.com/anatol/booster/issues/110 and potentially https://github.com/anatol/booster/issues/79 if there are no plans for a native UKI implementation

Zile995:
@c3Ls1US @anatol 

Or you can simply use the [booster-ukify](https://aur.archlinux.org/packages/booster-ukify) or [booster-um](https://aur.archlinux.org/packages/booster-um).

PR #254, opened 2024-01-27
Update documentation on UKIs

c3Ls1US:
- Corrects 'Universal Kernel Image' to 'Unified Kernel Image'


c3Ls1US:
@anatol part of the description for commit https://github.com/anatol/booster/commit/67eaed963ba20fc52eca4ead3a94ccf67738bd87 is wrong -- we don't write the output to `/boot/Linux/EFI`. 

Would this be a problem?

anatol:
> Would this be a problem?

The commit has landed master already. So it is too late to change it unfortunately.

PR #260, opened 2024-03-31
Add regenerate_uki script to Arch packaging

shvrma:
Even though systemd-ukify was added to optional package dependencies and the regenerate_uki script was provided - it was not included in the final package.

### DIFF
--- a/packaging/arch/PKGBUILD
+++ b/packaging/arch/PKGBUILD
@@ -60,6 +60,7 @@ package() {
   install -Dp -m644 docs/manpage.1 "$pkgdir/usr/share/man/man1/booster.1"
   install -Dp -m755 init/init "$pkgdir/usr/lib/booster/init"
   install -Dp -m755 packaging/arch/regenerate_images "$pkgdir/usr/lib/booster/regenerate_images"
+  install -Dp -m755 packaging/arch/regenerate_images "$pkgdir/usr/lib/booster/regenerate_uki"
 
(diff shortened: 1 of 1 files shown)


packaging/arch/PKGBUILD
```diff
@@ -60,6 +60,7 @@ package() {
   install -Dp -m644 docs/manpage.1 "$pkgdir/usr/share/man/man1/booster.1"
   install -Dp -m755 init/init "$pkgdir/usr/lib/booster/init"
   install -Dp -m755 packaging/arch/regenerate_images "$pkgdir/usr/lib/booster/regenerate_images"
+  install -Dp -m755 packaging/arch/regenerate_images "$pkgdir/usr/lib/booster/regenerate_uki"
```
anatol, line 63: You need to install `regenerate_uki` file here.

anatol (changes requested):

packaging/arch/PKGBUILD
anatol: ping?

shvrma: Sorry for not answering for the first time, the reason is that I didn't understand the question. Are you also asking to trigger the regeneration of all UKIs in pacman hooks? The reason for that pull request was that the `regenerate_uki` script was not included in my install and my investigation brought me here

anatol: You are installing **regenerate_images** as `regenerate_uki` and it is probably not what you really want.

PR #264, opened 2024-07-09
init: check hidraw descriptor to figure out fido or not

7Ji:
this checks the Usage Page value from report_descriptor against 0xd0f1, similar to how it's done in libfido2

references:
- https://github.com/Yubico/libfido2/blob/03c18d396eb209a42bbf62f5f4415203cba2fc50/src/hid_hidapi.c#L146
- https://www.usb.org/sites/default/files/hid1_11.pdf 6.2.2.7 Global Item

tested with pico-fido


anatol:
Thank you!

PR #267, opened 2024-08-02
Forward arguments set by the kernel cmdline to the init system

### LINKED ISSUE #266: Booster doesn't pass init process arguments properly
As title says. I discovered it when trying to pass the -q flag to dinit init system. Normally, the kernel command would look something like this: "kerneloptions initrd etc -- -q" the double dashes marking the beginning of arguments that are to be passed to the init process. On mkinitcpio, the flag is passed as expected and calling `ps -fp 1 `clearly lists its presence. On booster, it unfortunately doesn't happen even though when checking dmesg, the parameters passed to the init process are correctly listed.

Kuchteq:
Kernel can pass arguments to the init script if it encounters a [double dash ](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html) that should then be forwarded to the actual init system (at least that's the behavior on in other init scripts, such as ones generated by mkinitcpio).

Fixes #266

### DIFF
--- a/init/main.go
+++ b/init/main.go
@@ -666,6 +666,11 @@ func switchRoot() error {
 	}
 
 	initArgs := []string{initBinary}
+        // Kernel can pass their arguments to the init program that we should then forward to actual system init
+        if(len(os.Args) > 1) {
+                initArgs = append(initArgs, os.Args[1:]...)
+        }
+
 	isSystemdInit, err := isSystemd(initBinary)
 	if err != nil {
 		return err


anatol:
Thank you for the patch! It is useful to have such functionality.

It also needs
- documentation, I think https://github.com/anatol/booster/blob/master/docs/manpage.md#boot-time-kernel-parameters would be the best place to add information about parameters handling.
- tests to verify the functionality

thePrivacyFanatic:
this when completed would be nice for people looking for a silent boot, as init systems like Dinit take the args without looking for more args in /proc/cmdline.

maybe you would prefer to exclude [args used by booster](https://github.com/anatol/booster/blob/master/docs/manpage.md#boot-time-kernel-parameters) from being forwarded or do other inits also forward all args?

Issue #270, opened 2024-10-01
add gentoo kernel config path search


alexminder:
Hello,
On Gentoo kernel config placed in `/lib/modules/KVER/config`. Can you add it for booster in to search path?

anatol:
Are you talking about kernel config file?

If yes the code is located here https://github.com/anatol/booster/blob/276a3efc4336e866e3944359ecd86658d6354eb9/generator/kmod.go#L732

note that booster also falls back to `/proc/config` and `/proc/config.gz` files. Do you have it available at your machine?

alexminder:
> Are you talking about kernel config file?
> If yes the code is located here
> https://github.com/anatol/booster/blob/276a3efc4336e866e3944359ecd86658d6354eb9/generator/kmod.go#L732
Yes.

I added block wich works for me:
```go
      } else if f, err := os.Open("/lib/modules/" + kernelVersion + "/config"); err == nil {
              // Gentoo specific config path.
              debug("reading %s", f.Name())
              defer f.Close()
              r = f
```

> note that booster also falls back to `/proc/config` and `/proc/config.gz` files. Do you have it available at your machine?
Config in /proc shows current (booted) kernel version. But if we build initramfs for new kernel version with changed kernel config, a new kernel may fail to boot. For example, root device driver was built-in , but in new config it changed to module.

anatol:
@alexminder your change looks good to me. Could you please send a PR with it?

A related item is to add Gentoo into set of tested platforms. To make that happen I need a script that creates an image with Gentoo system. Something like this one https://github.com/anatol/booster/blob/master/tests/generators/alpinelinux.sh

anatol:
@alexminder the question with Gentoo testing is still open.

If you can provide me the simplest set of commands that creates a Gentoo image (similar to this script https://github.com/anatol/booster/blob/master/tests/generators/alpinelinux.sh) then I can add an integration test for Gentoo and it will help to catch problems with the disto early.

desultory:
> @alexminder your change looks good to me. Could you please send a PR with it?
> 
> A related item is to add Gentoo into set of tested platforms. To make that happen I need a script that creates an image with Gentoo system. Something like this one https://github.com/anatol/booster/blob/master/tests/generators/alpinelinux.sh AUR has `portage` package which is a good sight that creating a Gentoo image is possible.

If you'd like to support gentoo, installkernel hooks would be very nice.

PR #272, opened 2024-10-07
Add gentoo dist-kernel specific config path

alexminder:


alexminder:
Fixes issue #270

anatol:
Thank you!

PR #279, opened 2024-12-16
packaging: Add kernel-install hook for booster

tmccombs:
This allows booster to be used as the initrd_generator with kernel-install (1).

### DIFF
--- a/packaging/common/50-booster.install
+++ b/packaging/common/50-booster.install
@@ -1,0 +1,31 @@
+#!/usr/bin/bash -e
+
+COMMAND="${1:?}"
+KERNEL_VERSION="${2:?}"
+BOOT_DIR="$3"
+KERNEL_IMAGE="$4"
+
+# If the initrd was provide don the command line, we shouldn't generate our own
+# If the supplied image is a UKI, we don't need to create an initrd
+if [[ "$COMMAND" != "add" || "$#" -gt 4  || "$KERNEL_INSTALL_IMAGE_TYPE" = "uki" ]]; then
+    exit 0
+fi
+
+# Only create an initrd if we are the configured initrd generator
+if [[ "$KERNEL_INSTALL_INITRD_GENERATOR" != "booster" ]]; then
+    exit 0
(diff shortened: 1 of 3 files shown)


packaging/common/50-booster.install
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+# Only create an initrd if we are the configured initrd generator
+if [[ "$KERNEL_INSTALL_INITRD_GENERATOR" != "booster" ]]; then
+    exit 0
+fi
+
+existing_image="$KERNEL_IMAGE%/*}/initrd"
+output_path="$KERNEL_INSTALL_STAGING_AREA/initrd"
+if [[ -f "$existing_image" ]]; then
```
tmccombs, line 21: This was based on the dracut. I'm not entirely sure when there would be an initrd file next to the kernel.

I also wonder if, if there is an existing "booster-*.img" file for the current kernel in the boot directory, if it should just copy that instead of building a new image.

anatol: > This was based on the dracut. I'm not entirely sure when there would be an initrd file next to the kernel.

At Arch Linux booster generates a corresponding initramfs file for each installed kernel. And puts the this generated file next to the kernel binary.

> I also wonder if, if there is an existing "booster-*.img" file for the current kernel in the boot directory, if it should just copy that instead of building a new image.

I would say let's do what other tools do already.

anatol:
Any objections for this change? If not - I am going to merge it to the codebase.

ThirteenPeeps:
I think mkinitcpio and dracut include microcode in their generated initramfs, and I know booster purposely doesn't.

So presumably for mkinitcpio/dracut folks, microcode gets applied as expected, but using kernel-install with this booster hook will ignore microcode entirely.

Wondering if this hook could/should include something to make sure any ucode files get included in the bootloader entry? E.g. checking for `*-ucode.img` files in /boot and copying them to the output directory

anatol:
@Foxboron merged now

PR #285, opened 2025-02-16
ukify: Use --microcode instead of --initrd

technobirb:
Cosmetic. systemd [added](https://github.com/systemd/systemd/commit/d380337dc5141eeb3c25071c6fcd9337aa814b62) the [--microcode](https://man.archlinux.org/man/ukify.1#%5BUKI%5D_section) option. This is probably the same as putting the ucode img as the 1st initrd.


anatol:
thank you for the patch @technobirb !

cc @oech3 for additional review

thePrivacyFanatic:
in the time that passed the script was deprecated in favor of a kernel-install plugin (which has its own ways to handle UKIs) so I assume this matters less as of now

PR #287, opened 2025-03-15
Bugfix: close device file descriptors in wwid

basploeger:
File descriptors to block devices (like /dev/sda) currently remain open and are inherited down the userspace process tree through PID 1. For example, on my machine the LVM command `vgchange` issues this warning during system initialization:

```
File descriptor 6 (/dev/sda) leaked on vgchange invocation.
```

This change closes the file descriptors properly, eliminating those warnings from the boot messages.


anatol:
Fantastic, thank you for the fix!

PR #302, opened 2025-07-30
add noLogging level

Indithem:
I had a specific problem which logs `mount(/dev/nvme0n1p5): no such file or directory` just after bootloader. I used systemd's ukify to add a splash image. I also use ` quiet loglevel=3 rd.udev.log_level=3 splash` to ensure no message can be put.

The only message that passes through is from booster. This logging simply destroys the splash image. :<

I'm sure this error is meaningless, as I've checked after boot for any mount errors and didn't find any.

This patch should ideally disable any logging from booster.

Incase there are others that experience the same thing and are ok with no logging, I think this patch justifies them.


anatol:
Thank you. The change looks good, I am going to merge it. I would like to make a small cleanup to your PR and start enum numbering with `iota` instead of less traditional `iota - 1`.

anatol:
Merged as fd70557436ff01645c5454ec4773617ae306f40b

PR #314, opened 2026-02-15
Add Plymouth boot splash support

### LINKED ISSUE #44: Plymouth support
Plymouth is a project from Fedora and now listed among the freedesktop.org's official resources providing a flicker-free graphical boot process. It relies on kernel mode setting (KMS) to set the native resolution of the display as early as possible, then provides an eye-candy splash screen leading all the way up to the login manager.

Plymouth is used by many Linux distributions. These include Ubuntu, fedora and pop os to name a few. It would be very useful for down stream distro maintainers to be able to use booster along with plymouth.

pilotstew:
## Summary

- Closes #44
- Adds Plymouth integration for graphical boot splash with LUKS password prompts
- Generator bundles plymouthd, plugins, themes, fonts, DRM modules, and XKB keyboard data into the initramfs
- Init starts plymouthd after waiting for a DRM device, creates udev database entries for GPU detection, and routes LUKS password prompts through Plymouth's graphical dialog with console fallback
- Enabled via `enable_plymouth` config option plus `splash` kernel parameter

## How it works

**Generator** (`generator/generator.go`):
- Detects default Plymouth theme via `plymouth-set-default-theme`, with `details` fallback
- Parses theme `.plymouth` files for font requirements (Pango font descriptions)
- Bundles plugins, renderers, theme files, fonts, fontconfig, XKB keyboard data
- Detects host GPU modules from sysfs and includes them alongside simpledrm/drm/drm_kms_helper
- Creates DRM udev database entries so Plymouth's device initialization check passes

**Init** (`init/plymouth.go`):
- Waits for a DRM device (prefers real GPU over simpledrm, with timeout fallback)
- Creates `/run/udev/data/` entries for DRM devices (same pattern as `devMapperUpdateUdevDb`)
- Starts plymouthd with ping-based readiness detection
- LUKS password prompts route through `plymouth ask-for-password` with console fallback
- Proper lifecycle: `show-splash` → password prompts → `update-root-fs` → `quit`
- `waitForPlymouthInit()` synchronization prevents LUKS handlers from racing plymouthd startup

**Config** (`generator/config.go`, `init/config.go`):
- `enable_plymouth: true` in `/etc/booster.yaml`
- Requires `splash` on kernel command line (matches dracut/mkinitcpio convention)

## Test plan

- [x] Unit tests for `extractFontFamily()` — Pango font description parsing (13 cases)
- [x] Unit tests for `parseThemeFonts()` — theme file font extraction (5 cases)
- [x] Unit tests for `findFontFiles()` — font directory matching with normalization (7 cases)
- [x] `go vet ./...` passes for both `generator/` and `init/`
- [x] Tested on real hardware: LUKS-encrypted btrfs root, AMD GPU, Cantarell theme
- [x] Verified fallback: boots normally without `splash` param, falls back to console without DRM device

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ferrreo:
The plymouth lib dir on debian based systems is /usr/libexec/plymouth not /usr/lib/plymouth  for the escrow and /usr/lib/x86_64-linux-gnu/plymouth/ for the plugins so ideally this needs some sort of detection for all the various distro paths for this as it will error currently.

ferrreo:
Another issue, it isn't following the image path in themes like it does the fonts so themes error with missing images.

pilotstew:
> The plymouth lib dir on debian based systems is /usr/libexec/plymouth not /usr/lib/plymouth  for the escrow and /usr/lib/x86_64-linux-gnu/plymouth/ for the plugins so ideally this needs some sort of detection for all the various distro paths for this as it will error currently.

You know I almost included that but I wasn't sure booster was used outside arch based systems. I can make that change.

pilotstew:
The following table shows Plymouth's boot sequence across three hardware configurations.                                                                                                      
  The current implementation requires the GPU driver to be listed in `modules_force_load`.
                                                                                                                                                                                                
  | Phase | GPU force-loaded (`modules_force_load: amdgpu`) | simpledrm only (VM / embedded) | Real GPU — driver NOT force-loaded |                                                             
  |---|---|---|---|
  | Kernel init | simpledrm initializes | simpledrm initializes | simpledrm initializes |
  | Booster: module load | GPU driver loaded early | No GPU driver present | GPU driver not loaded |
  | `waitForDrmDevice()` | Real GPU card found | Times out, simpledrm fallback | Times out, simpledrm fallback |
  | `createDrmUdevEntries()` | Entries created for real GPU | Entries created for simpledrm | Entries created for simpledrm |
  | `plymouthd` starts | Starts on real GPU | Starts on simpledrm | Starts on simpledrm ⚠️ |
  | Splash / LUKS prompt | Works on real GPU | Works on simpledrm | Works on simpledrm |
  | `plymouthNewroot()` | Called before `switch_root` | Called before `switch_root` | Called before `switch_root` |
  | `switch_root` → systemd | Plymouth running on real GPU | Plymouth running on simpledrm | Plymouth running on simpledrm |
  | systemd: GPU driver loads | Already loaded, no-op | No GPU driver, simpledrm persists | udev loads GPU driver |
  | simpledrm teardown | N/A | N/A | Kernel destroys simpledrm card ⚠️ |
  | Plymouth renderer | Intact on real GPU | Intact on simpledrm | Rendering on dead device ⚠️ |
  | `plymouth-quit-wait.service` | Quits cleanly | Quits cleanly | `TimeoutSec=0` — hangs forever 🔴 |
  | **Boot result** | ✅ Login | ✅ Login | 🔴 Infinite spinner |

  **Root cause:** Booster does not run `udevd`. Plymouth's device monitor uses
  `udev_monitor_new_from_netlink(udev, "udev")` which only fires when udevd re-broadcasts
  events. Without udevd, Plymouth cannot detect the real GPU appearing after startup and
  cannot perform the simpledrm → GPU transition that dracut/mkinitcpio support natively.

  ---

  ## Resolution Options

  **Option 1: Exclude simpledrm from DRM device selection**
  Never allow Plymouth to start on simpledrm. If no real GPU is found, disable Plymouth entirely.
  - ✅ Eliminates the hang scenario
  - ❌ Breaks Plymouth on simpledrm-only systems (VMs, embedded, EFI framebuffer-only machines)

  **Option 2: Auto-detect GPU driver and add to `modules_force_load`**
  At image generation time, scan `/sys/class/drm/card*/device/driver` for non-simpledrm GPU
  drivers and automatically add them to `modules_force_load`.
  - ✅ Seamless — Plymouth always starts on the real GPU, no user configuration needed
  - ✅ simpledrm-only systems still work (no GPU detected, Plymouth starts on simpledrm and stays there)
  - ⚠️ User may not be aware their GPU driver is being force-loaded into the initramfs

  **Option 3: Validate and warn at generation time**
  At image generation time, detect non-simpledrm GPU drivers on the host. If `enable_plymouth`
  is set and the GPU driver is absent from `modules_force_load`, emit a build-time warning or error.
  - ✅ Preserves explicit user control over `modules_force_load`
  - ✅ Catches misconfiguration at build time rather than producing a silent hang at boot
  - ⚠️ User must act on the warning to fix their config
  
  
  ## Conclusion
  I've gone with option 3. In the previous commit this relied upon the user correctly configuring booster.yaml with little warning and could result in plymouth hang. The latest commit checks what GPU modules are available at img generation. If GPU modules are detected and not included in modules_force_load plymouth will be disabled with a warning.

anatol (approved):

PR #316, opened 2026-03-10
init: go-libfido2 integration via plugin

pilotstew:
Replaces the `fido2-assert` subprocess with native go-libfido2 integration via the Go `plugin` package, keeping the main init binary CGO-free.

## Why plugin model

The main init binary is built with `CGO_ENABLED=0` for static linking. go-libfido2 requires CGO. `fido2plugin.so` is built separately with CGO enabled and loaded at runtime via `plugin.Open` — only when `enable_fido2: true` is set in `booster.yaml`. Users without FIDO2 hardware pay no cost.

## Commits

1. `init: load go-libfido2 via plugin` — `fido2plugin.so` exports `Fido2Assertion`, `IsFido2PinInvalid`, `IsFido2PinAuthBlocked`; build-tag stubs (`fido2_cgo.go` / `fido2_nocgo.go`) satisfy the compiler for `CGO_ENABLED=0` builds
2. `init: replace fido2-assert with go-libfido2` — PIN routing via Plymouth or console, 3 retries, keyboard fallback when PIN is blocked; `fido2Mu` mutex serializes across concurrent LUKS goroutines
3. `init: token-timeout= and fido2-device=/tpm2-device= in rd.luks.options` — controls how long to wait for tokens before starting keyboard prompt; accept `fido2-device=`/`tpm2-device=` without error for systemd-cryptsetup compatibility
4. `generator: bundle fido2plugin.so when enable_fido2 is set` — new `enable_fido2` flag in `booster.yaml`; explicit opt-in needed for cmdline-configured LUKS since the generator doesn't parse kernel parameters
5. `tests: update TestSystemdFido2 for go-libfido2 plugin`

## Test plan

- [x] Unit tests: `cd init && go test -v`
- [x] QEMU integration: `cd tests && go test -v -run "TestSystemdFido2"`
- [x] Local: Yubikey (PIN + touch, passphrase fallback)

### DIFF
--- a/init/luks.go
+++ b/init/luks.go
@@ -131,6 +131,10 @@ func recoverFido2Password(devName string, credential string, salt string, relyin
 	fido2Mu.Lock()
 	defer fido2Mu.Unlock()
 
+	if plymouthEnabled {
+		plymouthMessage("") // clear "Waiting for FIDO2" now that device is detected and we have the lock
+	}
+
 	credID, err := base64.StdEncoding.DecodeString(credential)
 	if err != nil {
 		return nil, fmt.Errorf("invalid credential: %v", err)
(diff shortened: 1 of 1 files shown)


pilotstew:
Note on CGO_ENABLED requirement
                                                                                                                                                                                                
  The switch from fido2-assert subprocess to go-libfido2 introduces a CGO dependency. go-libfido2 wraps the native libfido2 C library via #cgo directives, so the init binary now requires CGO_ENABLED=1 to build (which is the default — this only affects anyone explicitly building with CGO_ENABLED=0).                                                                              
                            
  Previously, shelling out to fido2-assert meant the init binary had no C dependencies and could be built fully statically. The trade-off is intentional: direct library integration eliminates the subprocess timing races and path-resolution failures that made FIDO2 unlock unreliable in practice.
                                                                                                                                                                                                
  Happy to discuss if this is a concern.

PR #322, opened 2026-03-24
init: go-libfido2 integration via plugin

pilotstew:
Replaces the `fido2-assert` subprocess with native go-libfido2 integration via the Go `plugin` package, keeping the main init binary CGO-free.

## Why plugin model

The main init binary is built with `CGO_ENABLED=0` for static linking. go-libfido2 requires CGO. `fido2plugin.so` is built separately with CGO enabled and loaded at runtime via `plugin.Open` — only when `enable_fido2: true` is set in `booster.yaml`. Users without FIDO2 hardware pay no cost.

## Commits

1. `init: replace fido2-assert with native go-libfido2 plugin` — plugin loading via `fido2iface.Fido2Plugin` interface (single type assertion replaces three `Lookup()` calls); PIN routing via Plymouth or console, 3 retries, keyboard fallback when PIN is blocked; `fido2Mu` mutex serializes across concurrent LUKS goroutines; `token-timeout=` in `rd.luks.options`; accept `fido2-device=`/`tpm2-device=` without error for systemd-cryptsetup compatibility
2. `generator: add enable_fido2 config, bundle fido2plugin.so, deprecate fido2-assert` — new `enable_fido2` flag in `booster.yaml`; explicit opt-in needed for cmdline-configured LUKS since the generator doesn't parse kernel parameters; `extra_files: fido2-assert` treated as `enable_fido2: true` with deprecation warning for smooth upgrades; PKGBUILD and CHANGES updated
3. `tests: update FIDO2 tests for go-libfido2` — `enableFido2: true` replaces `extraFiles: fido2-assert`; PIN read from `BOOSTER_TEST_FIDO2_PIN` env var, test skips immediately if unset; `--crypttab /dev/null` passed in tests without crypttab to avoid EACCES

## Test plan

- [x] Unit tests: `cd init && go test -v`
- [x] QEMU integration: `cd tests && go test -v -run "TestSystemdFido2"`
- [x] Local: Yubikey (PIN + touch, passphrase fallback)

> This is a continuation of #316 (branch renamed from `libfido_crypttab` to `libfido2`).


tests/systemd_test.go
```diff
@@ -20,17 +20,17 @@ func TestSystemdFido2(t *testing.T) {
...
 	}
 	vm, err := buildVmInstance(t, Opts{
-		disk:       "assets/systemd-fido2.img",
-		kernelArgs: []string{"rd.luks.uuid=b12cbfef-da87-429f-ac96-7dda7232c189", "root=UUID=bb351f0d-07f2-4fe4-bc53-d6ae39fa1c23"},
-		params:     params,
-		extraFiles: "fido2-assert",
+		disk:        "assets/systemd-fido2.img",
+		kernelArgs:  []string{"rd.luks.uuid=b12cbfef-da87-429f-ac96-7dda7232c189", "root=UUID=bb351f0d-07f2-4fe4-bc53-d6ae39fa1c23"},
+		params:      params,
+		enableFido2: true,
```
anatol: It looks like this change requires a manual update from the user. Such breaking changes to the configuration need to be documents.

Could you please add an entry to the `CHANGES` file with information on *what need to be done by user*.

pilotstew:
> @pilotstew could you please push your changes here to this PR?

Sorry the branch names keep messing with me on this one.

anatol:
Merged. Great work! Thank you so much.

PR #330, opened 2026-04-13
crypttab: keyfile-on-device, header=, fido2-device=, tpm2-device=

pilotstew:
Extends `/etc/crypttab` support (built on #325) with the remaining advanced options. All new features are opt-in via crypttab entry options and require no configuration changes when not used.

### `keyfile=<path>` on a separate device
Keyfile path of the form `keydev:UUID=<uuid>` or `keydev:LABEL=<label>` causes booster to mount the named block device read-only, read the keyfile from it, then unmount. `keyfile-timeout=` controls how long to wait for the device to appear (default: `MountTimeout`).

### `header=`
Detached LUKS header support. Accepts the same `<path>`, `UUID=<uuid>`, or `LABEL=<label>` forms. When the header lives on a separate device, booster mounts it read-only under `/run/booster/hdrdev-<name>`.

### `fido2-device=auto` (requires #322)
Activates FIDO2 token unlock for the volume. Generator auto-detects this option and bundles `fido2plugin.so`. Init iterates attached FIDO2 devices, prompts for PIN (with Plymouth support), and falls back to keyboard passphrase after `token-timeout=` (default 30 s) if no device is present or all assertions fail.

### `tpm2-device=auto` (requires #322)
Activates TPM2 token unlock (existing behaviour, now also opt-in via crypttab).

### `token-timeout=<seconds>`
Overrides the per-entry timeout for hardware token operations (FIDO2/TPM2). Defaults to 30 s when `fido2-device=` or `tpm2-device=` is present.

## Implementation notes

- `statusMessage()` helper in `plymouth.go` routes to Plymouth or console transparently.
- `stopSeeding` channel cleans up the hidraw device seeder goroutine when FIDO2 succeeds.
- `errFido2FallbackToKeyboard` sentinel keeps fallback control flow explicit and avoids silent swallowing of errors.
- `isFido2PinBlockedError` interface method distinguishes "PIN blocked — needs reset" from "PIN invalid — retry".
- Generator `fido2-device=` auto-detection sets `enableFido2 = true` so `fido2plugin.so` is bundled without requiring explicit `modules_force_load`.

## Tests

- `TestCrypttabKeyfileDevice` — keyfile on separate ext4 image, QEMU
- `TestCrypttabHeader` — detached header on separate raw image, QEMU
- `TestCrypttabFido2` — full FIDO2 assertion via QEMU USB passthrough (skipped when no YubiKey PIN configured)
- `TestCrypttabFido2NoDevice` — fake systemd-fido2 LUKS token, exercises 30 s timeout → keyboard fallback path, no hardware required
- `TestCrypttabTPM2` — unchanged (pre-existing)
- `TestSystemdFido2` — existing test updated to use dynamic image creation to avoid loop-device/udev race

`tests/generators/systemd_fido2.sh` was rewritten to format and enroll directly on the image file (not via loop device) to eliminate the `cryptsetup`/`blkid` exclusive-lock conflict that caused intermittent `EBUSY` failures.

## Dependencies

- Standalone: keyfile-on-device and header= have no external dependencies beyond #325.
- fido2-device= and tpm2-device= require #322 (`pr/libfido2`) to be merged first.

Closes #319.

### DIFF
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -134,6 +134,13 @@ Some parts of booster boot functionality can be modified with kernel boot parame
 Booster supports unlocking LUKS volumes declared in `/etc/crypttab` (see **crypttab(5)**).
 Only entries marked with the `x-initrd.attach` option are bundled into the initramfs at image build time. `rd.luks.*` kernel parameters take precedence — if a cmdline parameter already covers a device, its crypttab entry is skipped.
 
+Booster-specific behaviour for selected options:
+ * **keyfile** `/path:UUID=xxx` (or `LABEL=`, `PARTUUID=`, `PARTLABEL=`) — keyfile on a separate device. Booster mounts the device read-only at boot, reads the key, then unmounts. The file is not bundled into the initramfs.
+ * **`header=`** — if the path is a plain absolute path the generator bundles the file into the initramfs automatically. The `/path:deviceref` form mounts the device at boot; `/dev/...` uses the raw block device directly.
+ * **`fido2-device=auto`** — when present the generator automatically bundles `fido2plugin.so`; no `enable_fido2: true` in the config file is required.
+ * **`tpm2-device=auto`** / **`fido2-device=auto`** — Booster defers the keyboard passphrase prompt until the token attempt fails or `token-timeout` elapses (default: wait indefinitely).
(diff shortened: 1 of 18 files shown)


pilotstew:
Closing to revise before submission.

PR #331, opened 2026-04-14
crypttab: keyfile-on-device, header=, fido2-device=, tpm2-device=

pilotstew:
Extends `/etc/crypttab` support (built on the now-merged crypttab-core and libfido2 work) with the remaining advanced options. All features are opt-in via crypttab entry options and require no configuration changes when not used.

## New options

**`keyfile=/path:UUID=xxx`** — keyfile on a separate block device. Booster waits for the device, mounts it read-only, reads the key, then unmounts. `keyfile-timeout=` controls how long to wait (default: `mount_timeout`).

**`header=`** — detached LUKS header. Plain absolute paths are bundled into the initramfs at build time. The `/path:UUID=xxx` form mounts the header device at runtime; `/dev/...` uses the raw block device directly.

**`fido2-device=auto`** — activates FIDO2 token unlock. The generator auto-detects this option and bundles `fido2plugin.so` without requiring `enable_fido2: true` in the config. Init iterates attached FIDO2 devices, prompts for PIN (with Plymouth support), and falls back to the keyboard passphrase after `token-timeout=` elapses (default: 30 s).

**`tpm2-device=auto`** — activates TPM2 token unlock with the same priority scheduling and timeout behaviour as `fido2-device=`.

**`token-timeout=`** / **`keyfile-timeout=`** — accept a bare integer (seconds) or any Go duration string (e.g. `30s`, `2m`).

## Plymouth improvements

The switch from the `fido2-assert` subprocess to the native `go-libfido2` plugin gives the init binary fine-grained control over the FIDO2 assertion flow for the first time. Under `fido2-assert`, the subprocess handled device interaction internally and returned only a final result — there was no opportunity to surface intermediate state like "waiting for touch" or "PIN incorrect, try again" to the user. With the plugin, each stage of the assertion is visible to the caller, making meaningful status messages possible.

This is why the Plymouth integration is appearing here rather than in the Plymouth PR: the messages it carries are only meaningful now. Routing them to either Plymouth or the console throughout the unlock flow prompted a small `statusMessage()` helper in `plymouth.go` that avoids repeated if/else blocks. The TPM2 PIN prompt was also updated to route through Plymouth consistently — a gap that had existed since TPM2 support was first added.

## Testing

`TestCrypttabKeyfileDevice` and `TestCrypttabHeader` use static assets and run entirely in QEMU.

`TestCrypttabFido2` requires a hardware FIDO2 device but not a specific one — any device recognised by `fido2-token -L` works. It is skipped when `BOOSTER_TEST_FIDO2_PIN` is unset. The recommended invocation to avoid exposing the PIN in shell history:

```zsh
read -s "pin?FIDO2 PIN: " && echo
BOOSTER_TEST_FIDO2_PIN="$pin" go test -v -run TestCrypttabFido2 . 2>&1 | tee /tmp/fido2-test.log
```

During image creation the PIN is written only to a tmpfs file under `XDG_RUNTIME_DIR`, passed to `systemd-cryptenroll` via `CREDENTIALS_DIRECTORY`, and deleted immediately after enrolment — it never touches disk.

`TestCrypttabFido2NoDevice` covers the token-timeout → passphrase fallback path without any hardware, using a static image with a fake `systemd-fido2` LUKS token. This was the primary development vehicle for iterating on the fallback logic.

`TestCrypttabTPM2` verifies `tpm2-device=auto` using the `swtpm` software emulator.

Closes #319.

### DIFF
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -134,6 +134,13 @@ Some parts of booster boot functionality can be modified with kernel boot parame
 Booster supports unlocking LUKS volumes declared in `/etc/crypttab` (see **crypttab(5)**).
 Only entries marked with the `x-initrd.attach` option are bundled into the initramfs at image build time. `rd.luks.*` kernel parameters take precedence — if a cmdline parameter already covers a device, its crypttab entry is skipped.
 
+Booster-specific behaviour for selected options:
+ * **keyfile** `/path:UUID=xxx` (or `LABEL=`, `PARTUUID=`, `PARTLABEL=`) — keyfile on a separate device. Booster mounts the device read-only at boot, reads the key, then unmounts. The file is not bundled into the initramfs.
+ * **`header=`** — if the path is a plain absolute path the generator bundles the file into the initramfs automatically. The `/path:deviceref` form mounts the device at boot; `/dev/...` uses the raw block device directly.
+ * **`fido2-device=auto`** — when present the generator automatically bundles `fido2plugin.so`; no `enable_fido2: true` in the config file is required.
+ * **`fido2-device=auto`** / **`tpm2-device=auto`** — Booster defers the keyboard passphrase prompt until the token attempt fails or `token-timeout=` elapses (default: 30 s).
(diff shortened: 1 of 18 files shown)


anatol:
It looks good, thank you!

PR #335, opened 2026-04-23
init: cache passphrases for multi-device LUKS unlock (closes #306)

### LINKED ISSUE #306: Unlock multiple LUKS volumes with same passphrase
Apologies if this is already supported, in that case I couldn't find any info about it

I'm on Arch and have two partitions, root and swap, encrypted with LUKS that share the same passphrase  
With mkinicpio, I just had to type the passphrase once, and it automatically used it for both disks, but using booster it asks for the passphrase twice  
It would be convenient to be able to tell booster to use a passphrase for multiple volumes

EDIT: I've found an actual issue with it: if I'm not fast enough in typing the password for swap, the system starts booting without decrypting it, which leads to a 1'30" timeout as systemd cannot find the decrypted volume, and obviously not having access to swap after that

pilotstew:
When multiple LUKS volumes share the same passphrase — the common
root+swap or btrfs RAID1 layout — booster previously prompted for each
device separately. This PR implements a passphrase cache so the user is
prompted exactly once.

**init/luks.go**
- `passphraseCache`: mutex-protected in-memory store of passphrases that
  successfully unlocked a volume during this boot
- `tryPassphraseAgainstSlots`: helper to try a passphrase against a
  device's key slots without going through the prompt loop
- `requestKeyboardPassword`: fast-path tries cached passphrases before
  prompting; stores passphrase in cache on success
- `keyboardMu`: global mutex serializing keyboard prompts across
  concurrent luksOpen goroutines — without this, the race between "check
  cache" and "store passphrase" means both goroutines see an empty cache
  and both prompt; with it, the second goroutine re-checks after the
  first completes and typically unlocks silently

The cache is intentionally scoped to keyboard-entered passphrases only.
Token secrets (FIDO2, TPM2, clevis) are device- or policy-specific and
not meaningful to share. Holding the passphrase as plaintext in RAM
during the initrd phase follows the same approach as systemd-cryptsetup
(which caches passphrases across crypttab entries by default) and
dracut's crypt module — both treat the initrd memory space as trusted,
since there is no swap, no other processes, and no persistent storage
written at that stage.

**Tests**
- `TestPassphraseCache`: two LUKS2 partitions, shared passphrase, root
  on partition 2; a second unanswered prompt prevents root from mounting,
  making cache failures reliably detectable as a timeout
- `TestLuksBtrfsRaid1`: two LUKS2-wrapped btrfs RAID1 members; both
  must unlock before btrfs assembles, making cache failures
  structurally impossible to hide

Closes #306

### DIFF
--- a/init/luks.go
+++ b/init/luks.go
@@ -65,6 +65,14 @@ func tryPassphraseAgainstSlots(volumes chan *luks.Volume, done <-chan struct{}, 
 	return false
 }
 
+// passphraseCache holds passwords that successfully unlocked a LUKS volume during
+// this boot, so subsequent volumes (e.g. btrfs RAID1 members) can be tried
+// automatically without prompting the user again.
+var passphraseCache struct {
+	sync.Mutex
+	passwords [][]byte
+}
+
 // rd luks options match systemd naming https://www.freedesktop.org/software/systemd/man/crypttab.html
 var rdLuksOptions = map[string]string{
 	"discard":                luks.FlagAllowDiscards,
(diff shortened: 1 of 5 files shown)


anatol (approved):

lucas-mior:
Sorry, it still does not work. I am using the latest booster version from github. My `booster.yaml` is empty. My systemd-boot config is:
```
title   Arch Linux (booster)
linux   /vmlinuz-linux
initrd  /booster-linux.img
options rd.luks.name=$LUKS1_UUID=root1
options rd.luks.name=$LUKS2_UUID=root2
options root=/dev/mapper/root1
options rootflags=subvol=@
options panic=30 kernel.sysrq=1 loglevel=6
options systemd.show_status=1
```

When booting:

```
Enter passphrase for root1: *
Unlocking...
```
And it just hangs.

PR #336, opened 2026-04-24
generator: include usbhid and hid_generic when FIDO2 is enabled

### LINKED ISSUE #277: Missing `usbhid` in Host-Specific Image
I have encountered an issue where the `usbhid` module is missing from the generated initramfs image when no USB devices are connected during the build process. This issue prevents the unlocking of LUKS volumes using FIDO2. It occurs specifically when creating a host-specific image, which only includes kernel modules used by the current host.

Force loading these modules (as in #236) is a potential solution, but it was removed in 31d06a47708a139c159cb96dc7dc067a64230db0.

pilotstew:
## Summary

- In host-specific mode, `usbhid` and `hid_generic` may be absent from the image if no USB HID device is connected at build time
- When a FIDO2 key is plugged in at boot, the kernel cannot load the driver, the udev bind event never fires, and `usbhidWg.Wait()` in `recoverFido2Password` blocks indefinitely
- Fix: when `conf.enableFido2` is true (set via `enable_fido2` in booster.yaml or auto-detected from a `fido2-device=` crypttab entry), activate `usbhid` and `hid_generic` with `filter=false` so they are unconditionally included in host-specific images

Force-loading is intentionally avoided. Boot opens the netlink socket inside the `udevListener` goroutine, then immediately calls `loadModules` on the main goroutine. If `usbhid` were force-loaded, it could bind to an already-connected key before the netlink socket exists, dropping the bind event. `usbhidWg.Wait()` would then block forever waiting for an event that already fired. The existing wait-group approach (added in 31d06a4) is correct — the modules just need to be present in the image.

Adds a generator unit test: places `usbhid` and `hid_generic` in the simulated modules directory with an empty `hostModules` list, enables FIDO2, and asserts both modules appear in the output image.

Closes #277.

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -292,6 +292,16 @@ func generateInitRamfs(conf *generatorConfig) error {
 		}
 	}
 
+	if conf.enableFido2 {
+		// FIDO2 security keys are USB HID devices. In host-specific mode,
+		// usbhid and hid_generic may be absent if no USB HID device is
+		// connected at build time. Include them unconditionally so the key
+		// is recognized when plugged in during the passphrase prompt.
+		if err := kmod.activateModules(false, false, "usbhid", "hid_generic"); err != nil {
+			return err
+		}
+	}
+
 	if err := kmod.resolveDependencies(); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


anatol (approved):

anatol:
Thank you again!

Issue #233, opened 2023-07-27
Unable to unlock root partition with tpm2 key


Prototik:
booster 0.11, both with and without universal:true.
[Dmesg and luksdump](https://gist.github.com/Prototik/8ac6fba29cd9226af280636eae422d54)

Main part of the log is
```
[    4.730062] booster: no tpm devices found after 3 seconds.
[    5.048481] booster: recovering systemd-tpm2 token #0 failed: clevis.go/tpm2: unable to load data: parameter 1, error code 0x1f : integrity check failed
```
Not sure why booster doesn't see tpm2, dracut works just fine with this setup.

anatol:
Thanks for reporting it.

Is it a regression from 0.10 booster version?

Also how did you enroll your your systemd-tpm2 partition? What parameters did you use? And what version of systemd do you have?

Prototik:
> systemd 254~rc3-1-arch

> systemd-cryptenroll --tpm2-device=auto --tpm2-pcrs=7+12 --wipe-slot=tpm2 nvme0n1p2

Not sure about regression from 0.10, I just give a try for booster. But I can try to downgrade and report result later if you wish.

Prototik:
The same issue with booster 0.10

Prototik:
Bisecting points me to c1b666775e9ac2bad4246078c8904c6a20976842

My assumption as tpm modules are built-in into the kernel in archlinux packages, booster doesn't stand a chance to capture associated udev event as tpm devices initialized even before booster enters it's main method. So we should check presence of /dev/tpmrm0 and don't use waiters if tpm already here.

Something like that:
```patch
--- a/init/udev.go
+++ b/init/udev.go
@@ -2,6 +2,9 @@ package main
 
 import (
 	"fmt"
+	"github.com/google/go-tpm/legacy/tpm2"
+	"io"
+	"net"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -73,8 +76,19 @@ var (
 )
 
 func udevListener() error {
-	// Initialize tpmReadyWg
-	tpmReadyWg.Add(1)
+	var dev io.ReadWriteCloser
+
+	if enableSwEmulator {
+		dev, _ = net.Dial("tcp", ":2321") // swtpm emulator is listening at port 2321
+	} else {
+		dev, _ = tpm2.OpenTPM("/dev/tpmrm0")
+	}
+	if dev != nil {
+		dev.Close()
+	} else {
+		// Initialize tpmReadyWg
[...]

PR #337, opened 2026-04-24
init: fix rd.luks.name suppressing crypttab fido2-device= and token unlock race

pilotstew:
Previously, when `rd.luks.*` kernel parameters and `/etc/crypttab` both described
the same device, the crypttab entry was silently discarded — kernel parameters
overrode crypttab entirely. This dropped security options like `fido2-device=` and
`tpm2-device=` that have no `rd.luks.*` equivalent. The fix merges rather than
discards: kernel parameters retain authority over device ref and mapping name,
while crypttab supplies token flags, keyfile, header, and other unlock options that
the cmdline cannot express.

This surfaced as a spurious passphrase prompt after a successful FIDO2 PIN+touch
unlock when `rd.luks.name=<UUID>=<name>` was present on the kernel cmdline
alongside `fido2-device=auto` in `/etc/crypttab`.

## Bug 1 — crypttab options silently dropped

When `rd.luks.name=<UUID>=<name>` is parsed, `findOrCreateLuksMapping` creates a
mapping with `tokenFido2=false`. The crypttab merge loop then saw
`luksMatchExists=true` for the same UUID and skipped the crypttab entry entirely,
dropping `fido2-device=auto` and any other crypttab unlock options.

With `tokenFido2=false`, `hasPriority=false`, so the FIDO2 token ran in the
non-priority branch which does not defer the keyboard prompt until after token
completion.

**Fix:** replace skip-if-exists with `mergeCrypttabOptions`, which copies token
flags, keyfile, keySlot, tries, header, tokenTimeout, and dm-crypt flags from the
crypttab entry into the cmdline-derived mapping. The cmdline's name and device ref
are preserved; crypttab augments the rest.

## Bug 2 — race between tokenWg.Done and close(done)

In the non-priority goroutine, `recoverTokenPassword` sends to `volumes` (triggering
the main thread's `close(done)`), returns, then `defer tokenWg.Done()` runs. These
two events are in different goroutines with no ordering guarantee: if `tokenWg.Done`
fires first, the waiter unblocks, sees `done` still open, and calls `startKeyboard`
— showing a passphrase prompt even though the volume is already unlocked.

**Fix:** call `closeDone.Do(func() { close(done) })` inside the non-priority
goroutine on success, before the deferred `tokenWg.Done` runs — matching the
priority branch and eliminating the race.

## Testing

- Unit tests in `init/crypttab_test.go` including `TestRdLuksNamePreservesCrypttabFido2`
  which reproduces the exact regression path
- Verified with a live boot: `rd.luks.name` present + `fido2-device=auto` in
  `/etc/crypttab`, FIDO2 PIN+touch unlock completes with no follow-on passphrase prompt

### DIFF
--- a/init/crypttab.go
+++ b/init/crypttab.go
@@ -176,15 +176,50 @@ func parseCrypttabDuration(s string) (time.Duration, error) {
 	return time.ParseDuration(s)
 }
 
-// luksMatchExists reports whether luksMappings already contains an entry for ref,
-// allowing rd.luks.* kernel parameters to take precedence over crypttab.
-func luksMatchExists(ref *deviceRef) bool {
+// findLuksMapping returns the existing luksMapping for ref, or nil if not found.
+func findLuksMapping(ref *deviceRef) *luksMapping {
 	for _, m := range luksMappings {
 		if deviceRefEqual(m.ref, ref) {
-			return true
+			return m
 		}
 	}
-	return false
+	return nil
+}
+
+// mergeCrypttabOptions merges security-relevant options from a crypttab entry (src)
+// into a cmdline-derived mapping (dst). dst's ref and name are preserved; crypttab
+// supplies token flags, keyfile, header, and other unlock options that rd.luks.*
+// params cannot express.
+func mergeCrypttabOptions(dst, src *luksMapping) {
+	if src.tokenFido2 {
+		dst.tokenFido2 = true
+	}
+	if src.tokenTpm2 {
+		dst.tokenTpm2 = true
+	}
+	// Adopt crypttab's token timeout when the cmdline mapping still has the
+	// default (30 s) and the crypttab entry carries an explicit value.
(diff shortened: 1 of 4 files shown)


anatol:
The bugs look independent. And in such case it is better to put them into separate git commits - this makes easier to handle the changes e.g. in case if one of them needs to be reverted.

pilotstew:
I'll split and recommit.

anatol:
Thank you @pilotstew for your awesome work!

pilotstew:
Done. There are a few other simple fix issues I'll try to tackle this week as well.

anatol:
Thank you again!

anatol:
> Done. There are a few other simple fix issues I'll try to tackle this week as well.

And if you are looking for other high-value issue then this one is a good candidate https://github.com/anatol/booster/issues/233. The goal of it is to bring `systemd-tpm` unlocker to be compatible with the recent systemd TPM changes.

pilotstew:
Sure, I'll take a look.

Issue #256, opened 2024-02-18
```vconsole: true``` causes systems with encrypted root to freeze.

### ISSUE THREAD

**Issue by CosmicFusion** [anatol/booster#256]
OS: PikaOS 3/Ubuntu 23.10
PC: intel iris laptop/nvidia pc/qemu VM
Kernel: 6.6.6/6.5.5/6.5.0

Expected Behavior: 
having ```vconsole: true``` in the booster configuration would cause an un-encrypted system to boot right up.

having ```vconsole: true``` in the booster configuration would cause an encrypted system to show unlock dialog with keymap from vconsole.

Current Behavior
having ```vconsole: true``` in the booster configuration would cause an un-encrypted system to boot right up.

having ```vconsole: false``` in the booster configuration would cause an un-encrypted system to boot right up.

having ```vconsole: false``` in the booster configuration would cause an encrypted system to show unlock dialog with US keymap.

having ```vconsole: true``` in the booster configuration would cause an encrypted system to freeze at console font loading never asking for luks password.

no font set in vconsole: 
![Screenshot from 2024-02-18 17-53-17](https://github.com/anatol/booster/assets/83735213/371aa79b-1ded-4ea4-8387-5b3baa61c850)

terminus font:
![Screenshot from 2024-02-18 00-45-39](https://github.com/anatol/booster/assets/83735213/5ac3c5d5-3e74-4e48-be87-ba50707c0f11)

vga font:
![Screenshot from 2024-02-18 00-48-50](https://github.com/anatol/booster/assets/83735213/7fb3d1f8-7b6c-4a85-ae86-c873a1bab5a2)

my /etc/booster.yaml
```
vconsole: true
extra_files: busybox
enable_lvm: true
modules_force_load: usbhid
```

my /etc/vconsole.conf
```
XKBMODEL="pc105"
XKBLAYOUT="us"
XKBVARIANT=""
XKBOPTIONS=""
BACKSPACE="guess"
KEYMAP=us
```

CosmicFusion:
Solved but not gonna Close


The issue turned out to be caused by faulty locale generation on my end.

Though it shouldn't be closed because it should show error message at runtime with debug on, or error at image regen time.

We only got it fixed by accident.

PR #338, opened 2026-04-24
vconsole: fix FONT_MAP/FONT_UNIMAP resolution and make setfont non-fatal

### LINKED ISSUE #207: image creating fails with vconsole set to true
The Arch booster version fails to create the image when trying to include the vconsole font settings using "vconsole: true" here at FONT_MAP it seems. The booster config here does not include any other entry. 

vconsole.conf has
```
KEYMAP=de-latin1-nodeadkeys                                                                
FONT_MAP=8859-2                                                                            
FONT=spleen-12x24
```
/usr/lib/booster/regenerate_images
2023/02/16 18:13:19 unable to find file for specified font '8859-2'

### LINKED ISSUE #234: `vconsole: true` blocks booting w/ booster exit status 71
(follow-up of #228, running on the same system

## Setup

* OS: Arch Linux, recently upgraded
* Kernel: Zen kernel 6.4.6.zen1-1 from Arch repo
* booster: 0.11-1 from Arch repo
* `/etc/booster.yaml`

```
modules_force_load: i915
modules: vmd,i915,intel_lpss_pci
extra_files: fido2-assert
vconsole: true
```

* Boot loader: systemd-boot with the following entry:

```
title Arch Zen
linux /vmlinuz-linux-zen
initrd /intel-ucode.img
initrd /booster-linux-zen.img
options rd.luks.name=<REDACTED UUID>=arch root=/dev/mapper/arch rootflags=subvol=@ rw add_efi_memmap random.trust_cpu=on intel_iommu=on
```

* Root Partition: Btrfs on a LUKS2 partition, with a `systemd-cryptenroll`-ed Yubikey 5 NFC 

## Expected Behavior

When the screen shows `Enter passphase for arch:`, with my Yubikey plugged in, the LED on the Yubikey should be flashing. Touching the Yubikey (the metal part, that is) should unlock the LUKS partition and continue the boot process

## Actual Behavior

The screen shows (The number in the square brackets may vary)

```
[    0.968319] booster: exit status 71:
Press ENTER to Reboot
```

Pressing Enter key on keyboard does nothing. Have to hard reboot / shutdown.

## Attempted Fiddling 1: Print Logs from Booster

* Add `booster.log=info,console` to kernel options
* Reboot

Result: Expected behavior happens (with extra logging printed). This looks like a Heisenbug.

## Attempted Fiddling 2: Print Logs from Kernel

* Switch to kernel linux-lts 6.1.39-1
* Remove `booster.log=info,console` from kernel options
* Reboot to this kernel

Result: the LTS kernel prints many lines of `[    <NUMBER>] blacklist: problem blackl` before `Enter passphrase for arch`. Then expected behavior happens. Works well.

pilotstew:
Two vconsole fixes: FONT_MAP and FONT_UNIMAP files were being looked up in the
wrong directory with the wrong suffixes, and a setfont failure was incorrectly
aborting the boot sequence.

## generator: fix vconsole FONT_MAP and FONT_UNIMAP file resolution

`FONT_MAP` entries (e.g. `8859-2`) were searched in `consolefonts/` using font
suffixes instead of `consoletrans/` using map suffixes — causing "unable to find
file" errors for any system with `FONT_MAP` set (closes #207). `FONT_UNIMAP` had
the same misdirection into `consolefonts/` rather than `unimaps/`.

Both lookups now mirror kbd's `kbdfile_find` resolution logic
(`src/libkfont/context.c`): each directory has its own suffix list
(`.trans`/`_to_uni.trans`/`.acm` for maps; `.uni`/`.sfm` for unimaps).
Gzip-compressed variants are transparently decompressed.

The generator now probes for the kbd base directory at startup (`/usr/share/kbd`
on Arch/Debian/openSUSE/Alpine; `/usr/lib/kbd` on Fedora/RHEL) rather than
hardcoding `/usr/share/kbd`.

Additionally fixes the config fields: `FONT_MAP` and `FONT_UNIMAP` were both
written to `FontFile` instead of `FontMapFile` and `FontUnicodeFile`, so `setfont`
never received them via `-m` and `-u`.

## init: make setfont failure non-fatal

`setfont` exits non-zero when the kernel (KMS) rejects a font as too large for
the framebuffer. This caused `vconsole: true` to abort the boot sequence before
the LUKS password prompt on affected systems (closes #234, #256, #290).

Font loading is cosmetic — boot now continues with a warning rather than crashing.

## Testing

- `TestVirtualConsoleFontMap`: regression test for the FONT_MAP/FONT_UNIMAP
  directory misdirection — verifies files land in the image and that
  `FontMapFile`/`FontUnicodeFile` are set correctly in the init config
- `TestConsoleSetFontFailureIsNonFatal`: verifies setfont failure does not
  propagate out of `consoleSetFont`
- Live boot verified on affected hardware

### DIFF
--- a/generator/console.go
+++ b/generator/console.go
@@ -13,8 +13,31 @@ import (
 	"strings"
 )
 
-// path to console fonts, adjust it to your distro (e.g. Fedora uses /usr/lib/kbd/consolefonts path for it)
-var consolefontsDir = "/usr/share/kbd/consolefonts/"
+// kbdBaseDir is the root of the kbd data tree. Arch/Debian/openSUSE/Alpine use
+// /usr/share/kbd; Fedora/RHEL use /usr/lib/kbd. Probe at startup so the generator
+// works on any distro without configuration.
+var kbdBaseDir = func() string {
+	for _, d := range []string{"/usr/share/kbd", "/usr/lib/kbd"} {
+		if _, err := os.Stat(filepath.Join(d, "consolefonts")); err == nil {
+			return d
+		}
+	}
+	return "/usr/share/kbd"
+}()
+
+var (
+	consolefontsDir = filepath.Join(kbdBaseDir, "consolefonts")
+	consoletransDir = filepath.Join(kbdBaseDir, "consoletrans")
+	unimapsDir      = filepath.Join(kbdBaseDir, "unimaps")
+)
+
+// Suffix lists mirror kbd's kfont_context defaults (src/libkfont/context.c).
+// kbdfile_find resolves a name by trying name+suffix for each suffix in order.
+var (
+	consolefontsSuffixes = []string{"", ".psfu", ".psf", ".cp", ".fnt"}
+	consoletransSuffixes = []string{"", ".trans", "_to_uni.trans", ".acm"}
+	unimapsSuffixes      = []string{"", ".uni", ".sfm"}
+)
 
 func (img *Image) enableVirtualConsole(vConsolePath, localePath string) (*VirtualConsole, error) {
 	debug("enabling virtual console")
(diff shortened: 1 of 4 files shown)


anatol (approved):

anatol:
Thank you for fixing it. it was one of those annoying issues.

PR #341, opened 2026-04-25
generator: deterministic CPIO image generation (closes #309)

### LINKED ISSUE #309: Undeterministic booster.alias causes different initramfs when it should be equal
```sh
booster build booster.img
booster build booster2.img

diff booster.img booster2.img  # they are different even though nothing has changed in the system

booster unpack booster.img bunpack
booster unpack booster2.img bunpack2

git diff bunpack bunpack2

```
The diff shows that the only difference is the file `usr/lib/modules/booster.alias`.
Sorting them makes the two images equal.
Is there a reason for that? I would like to be able to just diff them to check if they match or not.

pilotstew:
## Summary

Fixes two independent sources of non-determinism in image generation:

- **`booster.alias` ordering**: aliases for the same pattern mapping to different modules were emitted in randomized map iteration order. Now sorted by `(pattern, module)` before writing.
- **CPIO module ordering**: map iteration over `requiredModules` and goroutines racing to call `AppendContent` produced different CPIO entry orders across builds. Module names are now collected into sorted slices; goroutines decompress in parallel with results stored by index, then flushed to the CPIO in sorted order.

Also fixes a latent bug in the original code where a missing `return` after sending to `errCh` on a failed module path lookup allowed execution to continue with an empty path. The refactor to `readModuleContent()` eliminates this by design.

## Tests

- Updated `TestUniversalMode` and `TestHostMode` to assert sorted alias output directly (removing the workaround that manually sorted before comparing).
- Added `TestDeterministicImage`: builds the same image twice with 5 modules and overlapping aliases, asserts byte-for-byte equality. Verified to catch the pre-fix bug reliably.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


anatol (approved):

anatol:
Thank you!

PR #342, opened 2026-04-26
init: fix TPM2 integrity check failure for systemd v252+ enrolled tokens (closes #233)

### LINKED ISSUE #233: Unable to unlock root partition with tpm2 key

pilotstew:
Users who enrolled with `systemd-cryptenroll` v252 or later see:

```
recovering systemd-tpm2 token failed: clevis.go/tpm2: unable to load data: integrity check failed
```

Two changes in systemd broke compatibility:

**1. Persistent SRK (v252+)**

`systemd-cryptenroll` v252 began provisioning a persistent SRK at handle `0x81000001` via `tpm2_get_or_create_srk` and recording it in the LUKS2 token JSON as `tpm2_srk`. Booster was always deriving a transient primary via `CreatePrimary`. When a persistent SRK was used during enrollment, loading the sealed object against a freshly derived primary produces a different parent key, causing the integrity check to fail.

The fix avoids the `IESYS_RESOURCE` deserialization problem noted in #233 — we only need the `TPM2_HANDLE` embedded in the IESYS header (bytes 6–9, big-endian), which maps directly to a `tpmutil.Handle`. When `tpm2_srk` is present, `tpm2.Load` is called against the persistent handle. `FlushContext` is intentionally skipped on the persistent handle (it would evict it from the TPM). Pre-v252 tokens with no `tpm2_srk` continue to use the old `CreatePrimary` path.

**2. PBKDF2 PIN auth (v255+)**

`systemd-cryptenroll` v255 added a `tpm2_salt` field to PIN-protected tokens. The old formula was `authValue = SHA256_trimmed(pin)`; the new formula is `SHA256_trimmed(base64(PBKDF2-HMAC-SHA256(pin, salt, 10000, 32)))`. Tokens without `tpm2_salt` continue to use the old path.

**Coverage**

Four QEMU integration tests cover all relevant paths:

| Token format | Test |
|---|---|
| Pre-v252 (transient primary, no SRK) | `TestSystemdTPM2` (existing) |
| v252+ (persistent SRK, no PIN) | `TestSystemdTPM2SRK` (new) |
| v255+ (persistent SRK + PBKDF2 PIN) | `TestSystemdTPM2WithPin` (existing) |
| v252–254 (persistent SRK, SHA256 PIN, no PBKDF2) | `TestSystemdTPM2LegacyPin` (new) |

The legacy-PIN test uses raw `tpm2-tools` to construct the token independently of the installed systemd version and is skipped if `tpm2-tools` are absent.

**Known limitation**

The TPM2 PIN prompt does not yet include the mapping name and has no retry loop on wrong PIN. The FIDO2 path gained both (mapping name in prompt, 3-attempt retry with progressive feedback) in a recent PR; parity for TPM2 is planned as a follow-up.

### DIFF
--- a/init/luks.go
+++ b/init/luks.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"crypto/sha256"
 	"encoding/base64"
 	"encoding/binary"
 	"encoding/hex"
@@ -19,6 +18,7 @@ import (
 
 	"github.com/anatol/clevis.go"
 	"github.com/anatol/luks.go"
+	"github.com/google/go-tpm/tpmutil"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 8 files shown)


anatol (approved):

anatol:
Fantastic improvement! Thank you again.

Issue #203, opened 2023-01-19
Booster favors password / recovery key over Fido2 


0xSigi:
Hello, I am having issues booting with Booster that I haven't seen before.. Due to using Camellia as my cipher for Luks I am still using the wip branch (running Arch), today I've build a new machine and rolled everything as it was with my other rigs. however to my surprise booster favors recovery key (or a regular password) over my enrolled Fido2 key.. I did everything the same way as I did before so I am not sure from where this behavior comes.. To be able to use Fido I have to press enter a couple of times when being prompted for password and after few tries it asks me for my PIN.. This is rather cumbersome and tedious even thou I don't reboot that often.. 

Any idea what could be causing this? Once it asks for PIN it unlocks fine using yubikey.

Issue #198, opened 2022-12-06
Handle TPM2 + passphrase


codicodi:
A while ago systemd got support for TPM2 + passphrase setup:
https://github.com/systemd/systemd/pull/22563

It would be nice if booster supported unlocking such volumes as well

anatol:
I added booster boilerplate to handle PIN. Now I need to use the pincode to unseal the value.

I am looking for help from someone who understands the `tpm2-tss` API and can translate the systemd's change into `go-tpm` equivalent. See booster's `tpm2Unseal()` function and its new parameter `password` that is sha256 hashed content on PIN.

anatol:
Alright, I think I figured out the logic. I improved booster code and the test shows it is able to unlock a drive locked with systems tpm+pin.

Please pull the changes from `wip` branch and test it with your setup.

codicodi:
Thanks for looking into this. 
Unlocking does indeed work, but the whole procedure appears a bit messy. 
First, booster asks for a passphrase which I guess refers to a recovery key also enrolled on this drive. I entered it incorrectly and booster asked for a PIN next. I entered the PIN (correctly) and booster once again showed the passphrase prompt, but the system booted after a short while anyway without me entering anything.

Interestingly the log complains about no tpm devices being found just before obtaining password for the tpm2 token

![20221214_222225](https://user-images.githubusercontent.com/22731198/207726442-6d17a742-c8a7-4d6e-96b0-ec74b20aee9c.jpg)
[booster.log](https://github.com/anatol/booster/files/10232036/booster.log)

anatol:
Booster tries to unlock all possible LUKS slots in parallel. Such an algorithm would work well if slots are non-interactive (non-PIN tpm, clevis, ...). The first valid slot unlocks the partition, and other slots processors get canceled.

In your case, you have 2 interactive slots that require some sort of pin/password. So the prompts you see are interleaved.

My guess you expected to see only 1 interactive slot processed at a time. But which one should be processed/prompted first? And why?

codicodi:
A TPM2 PIN is supposed to be safe, yet convenient (lowish entropy; easy to type). If a user went to trouble of enrolling it, it's probably the preferable way of unlocking. 
Regular passphrases (including recovery key) could be tried next. Correct me if I'm wrong, but I think the order no longer matters since booster can just keep asking until given passphrase matches one of the slots.

PR #344, opened 2026-04-28
init: remove crypttab fido2-device=/tpm2-device= token flags in favour of LUKS2 header

pilotstew:
Booster already reads the LUKS2 header via `d.Tokens()` for every unlock attempt — the same metadata `systemd-cryptenroll` writes. The `fido2-device=` and `tpm2-device=` crypttab options were therefore redundant: they duplicated information already available in the header and inconsistently with every other token type (clevis, recovery) which never required crypttab annotation.

The flags also left a correctness gap. Without them, `luksOpen` had no `priorityTypes` entry for the volume, so the keyboard goroutine would try passphrase unlock against hardware-credential slots (TPM2, FIDO2, clevis) where a typed passphrase can never succeed.

**Changes:**

- `fido2-device=` and `tpm2-device=` are still accepted in crypttab for compatibility but have no effect
- `tokenFido2`/`tokenTpm2` fields and the `priorityTypes`/`hasPriority` mechanism removed from `luksMapping` and `luksOpen`
- Slot exclusion is now unconditional: the keyboard goroutine always skips slots claimed by any token in the LUKS2 header, with fallback to all slots only when no dedicated passphrase slot exists
- The 30s `token-timeout` default now applies to all LUKS entries, not just those with crypttab token flags — any LUKS2 volume may carry enrolled tokens

### DIFF
--- a/init/crypttab.go
+++ b/init/crypttab.go
@@ -134,9 +134,9 @@ func parseCrypttabReader(r io.Reader) ([]*luksMapping, error) {
 					m.header = hdrPath
 					m.headerDeviceRef = hdrRef
 				case strings.HasPrefix(opt, "fido2-device="):
-					m.tokenFido2 = true // value ("auto") ignored — booster auto-detects enrolled tokens
+					// accepted for compatibility; token detection uses LUKS2 header
 				case strings.HasPrefix(opt, "tpm2-device="):
-					m.tokenTpm2 = true // value ("auto") ignored — booster auto-detects enrolled tokens
+					// accepted for compatibility; token detection uses LUKS2 header
 				case strings.HasPrefix(opt, "token-timeout="):
 					d, err := parseTokenTimeout(opt[14:])
 					if err != nil {
@@ -188,15 +188,8 @@ func findLuksMapping(ref *deviceRef) *luksMapping {
 
 // mergeCrypttabOptions merges security-relevant options from a crypttab entry (src)
 // into a cmdline-derived mapping (dst). dst's ref and name are preserved; crypttab
-// supplies token flags, keyfile, header, and other unlock options that rd.luks.*
-// params cannot express.
+// supplies keyfile, header, and other unlock options that rd.luks.* params cannot express.
 func mergeCrypttabOptions(dst, src *luksMapping) {
-	if src.tokenFido2 {
-		dst.tokenFido2 = true
-	}
-	if src.tokenTpm2 {
-		dst.tokenTpm2 = true
-	}
 	// Adopt crypttab's token timeout when the cmdline mapping still has the
 	// default (30 s) and the crypttab entry carries an explicit value.
 	if src.tokenTimeout > 0 && src.tokenTimeout != dst.tokenTimeout {
--- a/init/crypttab.go
+++ b/init/crypttab.go
@@ -154,7 +154,7 @@ func parseCrypttabReader(r io.Reader) ([]*luksMapping, error) {
 			continue
 		}
 
-		if (m.tokenFido2 || m.tokenTpm2) && !tokenTimeoutExplicit {
+		if !tokenTimeoutExplicit {
 			m.tokenTimeout = 30 * time.Second // systemd default: wait 30s for tokens before also prompting keyboard
 		}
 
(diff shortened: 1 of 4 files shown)


init/crypttab.go
```diff
@@ -134,9 +134,9 @@ func parseCrypttabReader(r io.Reader) ([]*luksMapping, error) {
 					m.header = hdrPath
 					m.headerDeviceRef = hdrRef
 				case strings.HasPrefix(opt, "fido2-device="):
-					m.tokenFido2 = true // value ("auto") ignored — booster auto-detects enrolled tokens
+					// accepted for compatibility; token detection uses LUKS2 header
 				case strings.HasPrefix(opt, "tpm2-device="):
-					m.tokenTpm2 = true // value ("auto") ignored — booster auto-detects enrolled tokens
+					// accepted for compatibility; token detection uses LUKS2 header
```
anatol, line 139: Shall be some kind of warning or at least info message here?

pilotstew: We could but nothing is actually lost in the change. To walk through what tpm2-device/fido2-device=auto used to do for a user who had it:

  1. Slot exclusion — still happens, now unconditionally from the header. Same result.
  2. 30s token-timeout — still applied, now universally. Same result.
  3. Token unlock — always went through d.Tokens() and recoverTokenPassword. Unchanged.

  The flag was never needed for token detection — that always came from the header. It was only needed to activate slot exclusion and the timeout, which now apply to all LUKS volumes regardless. Users with the flag get identical behavior. Users without it now get the correct behavior they were missing.

  In short no degradation. The option is still "honored" in the sense that everything it used to enable still happens — just via the unconditional path rather than a flag gate. It's closer to how luks works — a marker that signals intent booster was already satisfying. A warning would imply something is wrong or missing, which isn't true. A manpage update is most definitely warranted. Apologies for the oversight

anatol: Ack, thank you for the clarification and for the documentation update

pilotstew:
Background: how token priority evolved in booster

  This PR is one step in a progression that started with a concrete boot-time problem and is working toward a cleaner general solution. Here's the full arc:

  The original state (no priority)

  Before ddd7e39, booster had no concept of token priority. All enrolled tokens and the keyboard passphrase goroutine ran truly in parallel with no coordination. This caused two visible problems:

  1. Spurious passphrase prompts (reported in #198, #203): a token goroutine (TPM2, FIDO2, clevis) would succeed and signal done, but if tokenWg.Done fired before the signal propagated, the keyboard goroutine would wake, see done still open, and launch a passphrase prompt on screen — even though the volume was already unlocked. Issue #198 captured this directly: "In your case, you have 2 interactive slots that require some sort of pin/password. So the prompts you see are interleaved."
  2. Keyboard tried hardware-credential slots: a typed passphrase was attempted against TPM2/FIDO2 slots where it can never succeed, wasting time and confusing users. This compounded the UX problem in #203, where the user had to press Enter several times through failed passphrase attempts before FIDO2 PIN was even offered.

  ddd7e39 — crypttab-gated priority

  Introduced tokenWg, priorityTypes, and hasPriority. When fido2-device= or tpm2-device= was present in crypttab, the keyboard goroutine deferred its prompt until hardware tokens finished (or token-timeout elapsed). Slot exclusion was also tied to this flag. This resolved the interleaving for users who explicitly annotated their crypttab entries — but left users who enrolled tokens via systemd-cryptenroll without those flags unprotected.

  145c97c — race fix for non-priority goroutines

  Even in the non-priority path, a race existed: a non-priority token could succeed and trigger close(done) via the volumes channel, then tokenWg.Done could fire before that close landed, waking the keyboard goroutine prematurely. Fixed by having the goroutine itself call closeDone.Do on success before returning — matching the priority branch's approach.

  This PR (#344) — remove crypttab flag dependency for correctness guarantees

  The two behaviors that fido2-device=/tpm2-device= gated — slot exclusion and the 30s timeout — are now unconditional. Booster always reads the LUKS2 header via d.Tokens(), so it always knows what slots are claimed by hardware tokens. There's no reason a user without the crypttab flag should have the keyboard try those slots or get no timeout protection. fido2-device=/tpm2-device= continue to be accepted silently for compatibility.

  Planned follow-on: header-based PIN priority

  The remaining gap: for volumes with PIN-requiring tokens (fido2-clientPin-required: true, tpm2-pin: true) the keyboard prompt still races the PIN prompt, tpm typically loads first and would preclude fido2. The fix is to scan the token payloads before spawning goroutines and give only PIN-requiring tokens priority similar to systemd token id mechanics (keyboard waits); touch-only FIDO2, PCR-only TPM2, and clevis _could_ fire in parallel.

PR #347, opened 2026-04-30
init: fix TPM2 PIN token unlock — empty PCR policy digest and missing retry/skip

pilotstew:
Two TPM2 PIN improvements.

## Bug — `policyPCRSession` corrupts policy digest for PCR-free tokens

When a token is enrolled without PCR binding (`systemd-cryptenroll --tpm2-pcrs=""`), the stored policy digest is computed from `PolicyPassword` only. `policyPCRSession` was calling `tpm2.PolicyPCR` unconditionally — even when `len(pcrs) == 0`. An empty PCR selection still mutates the running policy digest in the TPM session, so the computed digest diverged from the stored one and unseal failed regardless of whether the PIN was correct.

## Feature — TPM2 PIN retry loop and empty-input skip (parity with FIDO2)

FIDO2 PIN tokens already had 3-attempt retry with per-attempt feedback and an empty-input path to skip the token and fall through to the passphrase prompt. TPM2 now matches:

- 3-attempt loop with `"TPM2 PIN incorrect — "` prefix on subsequent prompts
- Empty PIN on first attempt skips the token immediately, falling through to the passphrase prompt
- Prompt updated from the generic `"Please enter TPM pin:"` to `"Enter TPM2 PIN for <name>:"` consistent with FIDO2

## Tests

- `TestSystemdTPM2NoPcrPin` — regression test for the PCR digest bug; image enrolled with `--tpm2-pcrs=""`
- `TestSystemdTPM2PinSkip` — empty PIN falls through to passphrase
- `TestSystemdTPM2PinRetry` — wrong PIN then correct PIN succeeds
- `TestSystemdTPM2PinExhausted` — 3 wrong PINs fall through to passphrase
- `systemd_tpm2.sh` gains `CRYPTENROLL_TPM2_PCRS` and `KEEP_PASSPHRASE_SLOT` env vars to support the new asset variants

### DIFF
--- a/init/luks.go
+++ b/init/luks.go
@@ -253,6 +253,7 @@ var fido2Mu sync.Mutex
 var errFido2Skipped = errors.New("FIDO2 skipped by user")
 var errFido2PinInvalid = errors.New("FIDO2 PIN invalid")
 var errFido2FallbackToKeyboard = errors.New("FIDO2 falling back to keyboard")
+var errTPM2Skipped = errors.New("TPM2 skipped by user")
 
 func recoverSystemdFido2Password(t luks.Token, mappingName string) ([]byte, error) {
 	var node struct {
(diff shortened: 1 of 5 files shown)


anatol:
Thank you @pilotstew !

PR #349, opened 2026-05-04
init/luks: minor cleanups in keyboard unlock path

pilotstew:
Two follow-up cleanups in the LUKS keyboard-unlock path. No functional changes.

  - Extract the duplicated passphrase-cache snapshot+iterate block in
    requestKeyboardPassword into a `tryCachedPassphrases` helper. The
    duplication emerged when the post-keyboardMu second snapshot was added
    (#306) — collapse both into one helper.
  - Drop the keyboardOnce sync.Once that guarded a closure with a single
    call site; inline the closure body.
    
    
    
  Both changes are pure refactors. The full test suite passes locally.


anatol:
Thank you!

PR #351, opened 2026-05-04
init/luks: improve FIDO2 unlock error handling and PIN-attempt accounting

pilotstew:
Four small behavioural improvements to FIDO2 unlock that today fail opaquely or mis-account PIN attempts, plus a consolidated test file. Each commit is self-contained.

### 1. Silent skip when a plugged-in key isn't enrolled here

When the user has multiple FIDO2 keys plugged in (common: YubiKey + backup), only one is enrolled with a given volume. Today every other key logs an opaque `libfido2 error 46` at info level as the unlock loop iterates through them — noise rather than signal.

Recognise `FIDO_ERR_NO_CREDENTIALS` (0x2e/46) and `FIDO_ERR_INVALID_CREDENTIAL` (0x22/34) at the plugin layer via a new `IsFido2WrongDevice` predicate, surface them to the unlock loop as `errFido2WrongDevice`, and skip the device with a debug-level log instead.

### 2. Re-prompt with PIN entry when the token requires one

When credential metadata says no PIN is required but the token has had one set since enrollment (firmware update, policy change), the device returns `FIDO_ERR_PIN_REQUIRED` and the unlock fails opaquely. The user sees a `libfido2 error 54` log and is dropped to the passphrase prompt even though entering their PIN would have unlocked it.

Recognise `FIDO_ERR_PIN_REQUIRED` (0x36/54) via a new `IsFido2PinRequired` predicate, surface it as `errFido2PinNeeded`, and on first hit flip the local `pinRequired` flag, re-extend `maxAttempts` to 3, and re-enter the loop without consuming an attempt so the user gets a PIN prompt for this token.

### 3. Count PIN_AUTH_INVALID and UV_INVALID as wrong-PIN

`IsFido2PinInvalid` currently only matches `FIDO_ERR_PIN_INVALID` (49). Two related codes for the same wrong-secret family fall through to the opaque-info path and don't get retried:

- `FIDO_ERR_PIN_AUTH_INVALID` (0x33 / 51) — wrong PIN auth attempt
- `FIDO_ERR_UV_INVALID`       (0x3f / 63) — wrong on-device user-verification (built-in fingerprint, etc)

Neither is mapped to a named go-libfido2 error in v1.5.3, so detect them via the generic `libfido2 error N` string from `errFromCode`'s default case, mirroring the existing `IsFido2PinBlocked` pattern.

### 4. Don't consume a PIN attempt on touch timeout

When a PIN-bearing FIDO2 token times out waiting for touch (no finger inside the device's UP window), libfido2 returns `FIDO_ERR_USER_ACTION_TIMEOUT` (0x2f/47) or `FIDO_ERR_ACTION_TIMEOUT` (0x3a/58). Today the loop treats this as a generic error and breaks, so the user loses 1 of their 3 PIN attempts for a fumbled tap even though their PIN was correct.

New `IsFido2TouchTimeout` predicate covers both codes; on PIN devices the loop re-prompts without consuming an attempt. Non-PIN devices fall through unchanged (their loop only runs once).

### Tests

`init/fido2plugin/errors_test.go` covers all four predicates against their mapped sentinels, the unmapped generic-string codes, `errors.Is` wrapping, and negatives.

### DIFF
--- a/init/fido2_cgo.go
+++ b/init/fido2_cgo.go
@@ -71,3 +71,11 @@ func isFido2PinBlockedError(err error) bool {
 	}
 	return fido2plugin.IsFido2PinBlocked(err)
 }
+
+func isFido2WrongDeviceError(err error) bool {
+	loadFido2Plugin()
+	if fido2plugin == nil {
+		return false
+	}
+	return fido2plugin.IsFido2WrongDevice(err)
+}
--- a/init/fido2_cgo.go
+++ b/init/fido2_cgo.go
@@ -87,3 +87,11 @@ func isFido2PinRequiredError(err error) bool {
 	}
 	return fido2plugin.IsFido2PinRequired(err)
 }
+
+func isFido2TouchTimeoutError(err error) bool {
+	loadFido2Plugin()
+	if fido2plugin == nil {
+		return false
+	}
+	return fido2plugin.IsFido2TouchTimeout(err)
+}
--- a/init/fido2_cgo.go
+++ b/init/fido2_cgo.go
@@ -79,3 +79,11 @@ func isFido2WrongDeviceError(err error) bool {
 	}
 	return fido2plugin.IsFido2WrongDevice(err)
 }
+
+func isFido2PinRequiredError(err error) bool {
+	loadFido2Plugin()
+	if fido2plugin == nil {
+		return false
+	}
+	return fido2plugin.IsFido2PinRequired(err)
+}
--- a/init/fido2_nocgo.go
+++ b/init/fido2_nocgo.go
@@ -19,3 +19,7 @@ func isFido2PinAuthBlockedError(err error) bool {
 func isFido2PinBlockedError(err error) bool {
 	return false
 }
+
+func isFido2WrongDeviceError(err error) bool {
+	return false
+}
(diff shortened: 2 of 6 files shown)


pilotstew:
Apologies for the force-push — if you'd already started reviewing the
original three commits.

The two new commits (056af77 and 36ec23d) cover two more FIDO2
error-handling cases:

- 056af77 extends `IsFido2PinInvalid` to also count `PIN_AUTH_INVALID`
  (51) and `UV_INVALID` (63) as wrong-PIN attempts. Today they fall
  through to the opaque-info path and don't get retried.
- 36ec23d adds `IsFido2TouchTimeout` and stops a touch timeout from
  burning one of the user's three PIN attempts on PIN-bearing tokens.

anatol (approved):

anatol:
Thank you @pilotstew for your contributions to the project!

pilotstew:
You're welcome. I didn't really initially intend all of this. But I've been enjoying it. Really I just wanted a nice plymouth interface. When I shared it it was more of a hey I made it work for me if it's helpful to anyone else. Locally I've finally reached what I set out to do. I've been working on carving it up into manageable pieces the last couple days but it's done in principle. Some minor messaging/display tweaks.

PR #352, opened 2026-05-04
init/luks: dedup the plymouth-or-console password prompt

pilotstew:
`recoverFido2Password`, `recoverSystemdTPM2Password`, and `requestKeyboardPassword` each open-code the same shape: try plymouth when enabled, log a warning and fall back to the console reader on failure.

Extract `askPasswordWithFallback` in `plymouth.go` and call it from all three sites. ~23 lines deleted, no behaviour change. `init` and `generator` tests pass.


anatol (approved):

anatol:
Thank you!

PR #354, opened 2026-05-05
init/luks: convert done channel to context.Context

### LINKED ISSUE #305: Ability to echo asterisks/dots when typing in decryption passphrase?
I know this is a minor nit but would it be possible to add support for being able to echo dots when typing in a luks decryption passphrase? My computer's keyboard can take a few seconds to fully connect on boot and before switching to booster it was helpful to be able to see the dots show up to know I had actually begun typing. I know the little _ flickers when typing keys but that is a lot less noticeable in general. I tried scanning through the docs and I don't see anything about being able to do that currently.

pilotstew:
### What this PR does

Pure mechanical conversion of five functions in the LUKS unlock paths from `done <-chan struct{}` to `context.Context`:

- `tryPassphraseAgainstSlots`
- `recoverTokenPassword`
- `recoverKeyfilePassword`
- `tryCachedPassphrases`
- `requestKeyboardPassword`

Plus `luksOpen`'s local cancellation channel: `done := make(chan struct{}) + sync.Once` becomes `ctx, cancel := context.WithCancel(context.Background()) + defer cancel()`. The test mirror in `init/token_orchestration_test.go`'s `runPinTokens` helper tracks the new contract.

Zero observable change. +54/−53 net.

### Why ctx specifically

`done` signals one thing (stop). `ctx` is Go's canonical goroutine-coordination object. The upcoming PRs need cancellation (cancellable keyboard reads, FIDO2-PIN and TPM2-PIN prompt cancellation); ctx's deadline support remains available if later work wants to express the existing `token-timeout=` / `keyfile-timeout=` features through it. Adopting the standard idiom now keeps the cancellation surface uniform instead of layering ad-hoc mechanisms.

### Cleanups that fall out

- Drops the local `sync.Once` around `close(done)` — `context.CancelFunc` is already idempotent.
- Drops the redundant explicit `cancel()` after the `<-volumes` drain — the deferred `cancel()` at the top covers it.

---

### Context: the bug this series fixes

When the cooked-mode keyboard reader is blocked in `read(2)` and a non-interactive token (TPM2 PCR-only, touchless FIDO2, or clevis) wins the unlock race, the keyboard prompt has no way to honor the cancellation: the prompt stays on screen, `inputMutex` / `keyboardMu` stay held, and any subsequent volumes that need keyboard entry block until the user manually presses Enter on the now-pointless prompt. A long-standing dangling-prompt bug that has lived alongside booster's concurrent unlock paths.

This series of PRs fixes that bug and, in the process, brings the keyboard and Plymouth prompt UX into the same concurrency model as the rest of booster's unlock pipeline.

### Companion work already shipped

Recent merged PRs in the prompt and unlock space:

- **#314** — Plymouth boot splash support (the UX layer PRs 4–5 of this series extend).
- **#335** — cache passphrases for multi-device LUKS unlock (single-prompt UX for shared passphrases).
- **#337** — fix `rd.luks.name` suppressing crypttab `fido2-device=` and token unlock race.
- **#347** — TPM2 PIN unlock: 3-attempt retry and empty-PIN skip (PIN-prompt UX consistency).
- **#350** — deterministic token dispatch order (predictable prompt order across boots).
- **#353** — serial PIN-token dispatch (predictable ordering when multiple PIN tokens are enrolled).

This series adds the missing piece: prompts that notice when a sibling unlock has succeeded and get out of the way.

### Upcoming UX plan

A focused series of PRs to fix the bug and complete the prompt UX:

1. **This PR — `done` → `context.Context` in unlock paths.** Pure mechanical, zero observable change. Foundation: ctx becomes the cancellation idiom the rest of the work threads through.
2. **Raw-mode keyboard reader.** Replaces the cooked-termios reader with one that honors ctx cancellation. Keyboard prompt auto-dismisses within ~100 ms of a sibling unlock winning. Also echoes asterisks during passphrase entry (closes #305), fixes Ctrl+S freezing the prompt, and supports paste from password managers (relevant to #320).
3. **Extend cancellation to PIN prompts.** Threads ctx into `recoverFido2Password` / `recoverSystemdFido2Password` / `recoverSystemdTPM2Password`. Closes the console side: every prompt cancels on autounlock win.
4. **Plymouth direct-socket IPC.** Replaces exec-based `plymouth` CLI calls with direct socket I/O. Mechanical foundation for splash-side concurrency-aware UX. No user-visible change. Prompt cancel wired for https://gitlab.freedesktop.org/plymouth/plymouth/-/merge_requests/393
5. **Plymouth state-transition messaging.** Splash messages reflect concurrent unlock state ("Waiting for FIDO2 ...", "FIDO2 PIN incorrect", etc.).

### DIFF
--- a/init/luks.go
+++ b/init/luks.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"encoding/base64"
 	"encoding/binary"
 	"encoding/hex"
@@ -45,7 +46,7 @@ type luksMapping struct {
 
 // tryPassphraseAgainstSlots tries password against each slot, sending the opened
 // volume on volumes if successful.  Returns true on success.
-func tryPassphraseAgainstSlots(volumes chan *luks.Volume, done <-chan struct{}, d luks.Device, checkSlots []int, password []byte) bool {
+func tryPassphraseAgainstSlots(ctx context.Context, volumes chan *luks.Volume, d luks.Device, checkSlots []int, password []byte) bool {
 	for _, s := range checkSlots {
 		v, err := d.UnsealVolume(s, password)
 		if err == luks.ErrPassphraseDoesNotMatch {
@@ -56,7 +57,7 @@ func tryPassphraseAgainstSlots(volumes chan *luks.Volume, done <-chan struct{}, 
 		}
 		select {
 		case volumes <- v:
-		case <-done:
+		case <-ctx.Done():
 		}
 		return true
 	}
@@ -506,7 +507,7 @@ func tokenNeedsPin(t luks.Token) bool {
 	return false
 }
 
-func recoverTokenPassword(volumes chan *luks.Volume, done <-chan struct{}, d luks.Device, t luks.Token, mappingName string) bool {
+func recoverTokenPassword(ctx context.Context, volumes chan *luks.Volume, d luks.Device, t luks.Token, mappingName string) bool {
 	var password []byte
 	var err error
 
@@ -534,7 +535,7 @@ func recoverTokenPassword(volumes chan *luks.Volume, done <-chan struct{}, d luk
 	}
 
 	info("recovered password from %s token #%d", t.Type, t.ID)
-	return tryPassphraseAgainstSlots(volumes, done, d, t.Slots, password)
+	return tryPassphraseAgainstSlots(ctx, volumes, d, t.Slots, password)
 }
 
 // readKeyfile reads a keyfile at path, skipping offset bytes and reading at most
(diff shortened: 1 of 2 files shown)


init/luks.go
```diff
@@ -789,8 +790,8 @@ func luksOpen(dev string, mapping *luksMapping) error {
 	}
 
 	volumes := make(chan *luks.Volume)
-	done := make(chan struct{})
-	var closeDone sync.Once
+	ctx, cancel := context.WithCancel(context.Background())
```
anatol, line 793: Looks good. Thank you for this cleanup. It is what I wanted to do for a while.

Eventually there would be a more widespread use of context, and things like timeout should use contexts (e.g. here https://github.com/anatol/booster/blob/master/init/main.go#L1133)

PR #355, opened 2026-05-06
init/console: replace cooked-mode reader with raw-mode CSI scanner

### LINKED ISSUE #305: Ability to echo asterisks/dots when typing in decryption passphrase?

pilotstew:
### What this PR does

Replaces `console.go`'s cooked-termios reader (`readPasswordLine` + `readPasswordLocked` + `readPassword`) with a new `console_input.go` raw-mode reader that honors ctx cancellation. Builds on #354 which established `context.Context` as the cancellation idiom in unlock paths.

Three commits: implementation, scanner + pty-harness tests, manpage section. `console_input.go` is intentionally over-commented (state-machine flow diagram, per-flag termios rationale, acronym key block) to keep the dense FSM code review-tractable.

### Headline behaviour fix

When a non-interactive token (TPM2 PCR-only, touchless FIDO2, or clevis) wins the unlock race while the keyboard prompt is showing, the cooked-mode reader has no way to honor the cancellation: it stays blocked in `read(2)`. The prompt dangles, `inputMutex` / `keyboardMu` stay held, and any subsequent volumes that need keyboard entry block until the user manually presses Enter on the now-pointless prompt.

After this change the reader checks `ctx.Done()` between `Poll(100ms)` cycles, returns `ctx.Err()` within ~100 ms of cancellation, restores termios via defer, ends the prompt line cleanly, and releases the locks. End-to-end behaviour: when a token wins, the keyboard prompt dismisses on its own.

### Bonuses that fall out of raw mode

- **Asterisks echo as the user types** (closes #305).
- **Editing keys**: Backspace / Delete, Ctrl+W (word kill), Ctrl+U (line kill), Tab (mask toggle), Enter, Ctrl+C, Ctrl+D (EOF on empty buffer; no-op otherwise).
- **Bracketed-paste support**: terminal emulators bracket pasted text with `\x1b[200~ ... \x1b[201~` markers. The cooked-mode reader treated those as literal password bytes and corrupted entries from password managers. The new scanner consumes the markers and preserves the pasted content verbatim. Particularly relevant to the in-flight remote SSH unlock work in #320.

### Scope and acknowledged gap

Keyboard-passphrase path only. `askPasswordWithFallback` / `readPassword` / `readPasswordLocked` take `ctx context.Context` and the call site in `requestKeyboardPassword` threads it through (already ctx-aware after #354).

**FIDO2-PIN and TPM2-PIN call sites pass `context.Background()`** and continue to dangle on autounlock — closing this gap belongs to an upcoming PR that threads ctx into `recoverFido2Password` / `recoverSystemdFido2Password` / `recoverSystemdTPM2Password`. Honest about the gap so the staged review works.

### Scanner design

Per-byte FSM consuming ANSI CSI / SS3 / OSC / DCS sequences and bracketed-paste markers so they never enter the password buffer. UTF-8 multi-byte reassembly across read boundaries with overlong / surrogate / >U+10FFFF rejection (via `utf8.Valid`). Bounded CSI parameter accumulation (16 bytes) — garbage input can't lock up the scanner.

Inspired by Plymouth's `on_key_event` in `src/libply-splash-core/ply-keyboard.c`, with intentional divergences:
- Kept ISIG so Ctrl+C still cancels (Plymouth's `cfmakeraw` clears it).
- Accept both BS (`\x08`) and DEL (`\x7f`) as backspace; Plymouth only DEL.
- Proper Ctrl+W word-kill; Plymouth collapses Ctrl+W to line-kill.
- OSC, DCS, and bracketed-paste handling; Plymouth doesn't have them.
- UTF-8 validation; Plymouth doesn't validate.
- Bounded CSI parameter accumulation; Plymouth unbounded.

UTF-8 helpers (`trimLastCodepoint`, byte-pattern table) are direct ports of Plymouth's `ply_utf8_*` functions; the CSI scanner itself is reimplemented as a per-byte FSM rather than Plymouth's slice-scan.

### No new dependency

The scanner is stdlib Go only. Surveyed alternatives (`golang.org/x/term`, `charmbracelet/x/input`, `peterh/liner`, `eiannone/keyboard`, `mattn/go-tty`) either don't support cancellable raw reads (the bug we're fixing) or pull substantial transitive deps inappropriate for an initramfs.

### Termios contract

Raw mode clears: ECHO, ICANON, IEXTEN, IXON, IXOFF, BRKINT, INPCK, ISTRIP, INLCR, IGNCR. Keeps ISIG on. VMIN=1, VTIME=0. Every return path restores the original termios via defer.

No signal-based safety net (sigaction). A process killed mid-read could leave the terminal in raw mode until systemd takes over — acceptable for booster's PID-1 use case (rarely takes signals during boot).

### Testing

Two test files. Scanner unit tests in `console_input_test.go` exercise the state machine directly via `Feed()` (CSI / SS3 / OSC / DCS / bracketed-paste / UTF-8 split across reads / editing keys / killWord boundary semantics).

Pty-harness end-to-end tests in `console_input_pty_test.go` exercise ctx cancellation against a real Linux pty pair (`/dev/ptmx`):
- **Headline test:** ctx cancellation returns within ~100 ms with termios restored.
- Reads typed bytes; partial-input cancel discards buffer; **Backspace deletes one full UTF-8 codepoint not one byte**; empty password (Enter on empty); EOF mid-read returns no error.

Live-tested on local tty1: passphrase entry, FIDO2-touchless-wins-while-prompting (the headline fix verified), wrong passphrase retry, Tab masking toggle. Multi-volume and serial console behavior matches in design but not separately tested.

### Risks

- Critical boot path: bug here can prevent disk unlock.
- Dense state-machine code in a critical path — review burden is real. The file leans on extensive comments (state-machine flow diagram, per-flag termios rationale, acronym key block) to make it tractable.
- Termios restoration relies on Go defer; no sigaction safety net (noted above).
- Keyboard path only; PIN-prompt cancellation lands in an upcoming PR.

### Manpage

Adds a "Password entry" section under NOTES documenting editing keys, asterisk echo, auto-dismiss on autounlock-win, and language support.

### DIFF
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -181,6 +181,16 @@ UUID parameter can optionally be enclosed with quote symbol `"` though it is not
 `root=UUID=ac8299a8-91ce-4bf6-a524-55a62844b787`, `root=UUID="ac8299a8-91ce-4bf6-a524-55a62844b787"` (not recommended),
 `rd.luks.uuid=ac8299a8-91ce-4bf6-a524-55a62844b787`, `rd.luks.uuid="ac8299a8-91ce-4bf6-a524-55a62844b787"` (not recommended).
 
+### Password entry
+Editing keys beyond the standard:
+ * **Ctrl+W** — erase the previous word.
+ * **Ctrl+U** — erase the entire entry.
+ * **Tab** — toggle visibility of the typed characters (asterisks ↔ literal).
+
+When any unlock method wins the race (a TPM2 PCR-only or FIDO2 touchless token completes, or a sibling volume's keyfile reads successfully), the keyboard prompt auto-dismisses and boot continues.
+
+PIN prompts (TPM2-PIN, FIDO2-PIN) accept up to 3 attempts; press Enter empty PIN to skip the token to the next unlock prompt.
+
 ### Modules selection
 It is a note to summarize the algorithm that computes what modules are going to end up in the generated booster image.
 Initial module list for booster is `defaultModulesList` - a set of predefined hard-coded modules defined at `generator.go`.
--- a/init/console.go
+++ b/init/console.go
@@ -3,13 +3,10 @@ package main
 import (
 	"bytes"
 	"fmt"
-	"io"
 	"os"
 	"os/exec"
-	"sync"
 	"unsafe"
 
-	"github.com/anatol/booster/init/quirk"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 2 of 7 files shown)


anatol:
Fantastic! Thank you very much!

PR #357, opened 2026-05-06
init: prompt-aware status-message redraw and token-unlock confirmation

pilotstew:
### What this PR does

Two changes that together close out the console UX for booster's concurrent unlock pipeline:

**1. Prompt-aware `statusMessage` redraw.** When `statusMessage` fires during an active console password prompt, the current line is erased, the message prints, and the prompt is reprinted below — cursor stays at the bottom, asterisk count preserved. Without this a status message printed mid-prompt tears through the prompt line.

The new `promptVolumeUnlocked` helper lets `statusMessage` skip the redraw when the prompt's volume is already unlocked (its `ctx.Done()` has fired), avoiding reprinting a stale prompt that the cancellation logic hasn't yet torn down.

`readPasswordOn` now sets `consolePrompt.{active, text, done}` during each prompt — fields declared in #355 that become load-bearing only now that `statusMessage` consumes them.

**2. Token-unlock confirmation.** `recoverTokenPassword` fires `statusMessageTimed("X unlocked via Y", 3s)` on success. After #355 (keyboard prompt cancellation) and #356 (PIN-prompt cancellation) cleanly dismiss prompts when a sibling token wins the race, the user previously saw nothing telling them what happened — boot just continued. This adds the missing confirmation: "cryptroot unlocked via FIDO2", "swap unlocked via TPM2", etc.

`tokenFriendlyName` provides the short label per token type. `statusMessageTimed` clears the message after 3 seconds so it doesn't linger on the splash.

### Scope

Console UX only. Plymouth users get these messages too (`statusMessage` routes to plymouthd when enabled), but Plymouth-side polish — splash banner styling, native timed-message support, dismissing prompts on socket hangup — is separate work upstream of plymouthd.

### Testing

- `TestPromptVolumeUnlocked`: covers all three observable states of the active-prompt's `ctx.Done()` — nil, open, closed.
- `TestTokenFriendlyName`: pins down the labels in the unlock-confirmation message.
- The redraw protocol's visual correctness (line-erase + reprint with preserved asterisk count) is covered by live test rather than unit test — the relevant output goes through `fmt.Print`, which the pty harness can't capture.

Live-tested on local tty1: token-unlock confirmation appears after autounlock; status messages don't tear through prompts during PIN-incorrect retry.

### DIFF
--- a/init/console_input.go
+++ b/init/console_input.go
@@ -494,11 +494,10 @@ var consoleMu sync.Mutex
 //     it to skip a redraw once the volume is unlocked by another method
 //     (avoids redrawing a prompt that's about to be dismissed)
 var consolePrompt struct {
+	active    bool
+	text      string
 	asterisks int
-	// Foreshadowed for the upcoming Plymouth splash UX work — see comment above.
-	// active bool
-	// text   string
-	// done   <-chan struct{}
+	done      <-chan struct{}
 }
 
 // consolePrint writes msg without acquiring consoleMu. Callers must hold consoleMu.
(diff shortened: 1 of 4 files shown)


pilotstew:
Testing messaging can be hard — would love some more eyes here, likely rough edges in the wording and timing. The strings ("X unlocked via Y"), the 3-second timed-clear duration, and the choice between timed-clear vs ctx-driven clearing are all judgment calls. Would love feedback on anything that reads off.

anatol:
The idea and the code that implements looks good to me. It is a type of a change that improves usability of booster. And it would be good to have it. Thank you for the PR!

Issue #126, opened 2022-01-22
Provided key already used in mapping


ethindp:
For some reason I've started seeing this error occurring whenever I run the regeneration script or when I perform a system update. I'm not sure what's causing it and so when the regeneration fails pacman always falls back to mkinitcpio. I couldn't find any references about this problem... How do you resolve it?

anatol:
Could you please provide exact command you are running and its error message?

ethindp:
I'm running the standard regenerate_images command. The error is: 2022/01/27 20:02:49 provided key already used in mapping system76_acpi->kernel/drivers/platform/x86/system76_acpi.ko.zst

anatol:
This error means there are 2 modules with the same name. Try booster from `master` branch, it will give you a cleaner error message (reference https://github.com/anatol/booster/commit/20da1446fb3cde7acb01789d535a705817d5552e).

PR #358, opened 2026-05-07
init/plymouth: replace plymouth CLI with direct socket IPC

pilotstew:
## Summary

Talk to plymouthd directly over its abstract Unix socket
(`\x00/org/freedesktop/plymouthd`) instead of fork-exec'ing the
`plymouth` CLI client. The wire protocol is documented in Plymouth's
`ply-boot-protocol.h`.

Carved into 5 commits for review:

1. **`init/plymouth: replace plymouth CLI with direct socket IPC`** —
   add `init/plysocket.go` (transport), swap the 5 exec call sites for
   ping / show-splash / display-message / quit / update-root-fs. Drop
   `/usr/bin/plymouth` from the initramfs (~80KB binary plus
   `glib` + `libply-boot-client` transitive deps). No call-site
   signature changes; `plymouthAskPassword` stays on the exec path
   for now and converts in commit 5 when ctx threading lands.

2. **`init/plymouth: redirect plymouthd stderr to /dev/kmsg`** —
   inherited fds are closed (FD_CLOEXEC) when booster exec's to
   systemd, so a plymouthd that inherits booster's stderr receives
   EPIPE on its next stderr write and dies before
   `plymouth-start.service` can attach. Routing to `/dev/kmsg`
   survives the handoff and lands the diagnostic output in the
   kernel ring buffer.

3. **`init/plymouth: make plymouthMessage fire-and-forget`** —
   plymouthd is single-threaded; a synchronous in-process
   `display-message` call would block on splash state, slowing
   concurrent unlock work. Wrap in a goroutine.

4. **`init/plymouth: thread ctx through waitForPlymouthInit`** —
   small signature change so unlock paths can bail when a sibling
   token has already cleared the volume, instead of blocking on
   plymouthd init for a volume that's already being unlocked.

5. **`init/plymouth: ctx-aware password prompt with cancel-on-hangup`**
   (the headline) — `plymouthAskPassword(ctx, prompt)` plus a
   serialization mutex. On ctx cancel the underlying socket is closed;
   the goroutine returns cleanly. plymouthd builds whose
   connection-hangup handler tears down pending prompts also dismiss
   the on-screen UI on close — older builds leave the UI visible
   until the splash is otherwise cleared, but boot proceeds correctly
   either way (matching upstream's prior behaviour minus the orphaned
   plymouth subprocess and leaked goroutine the exec path produced).
   `askPasswordWithFallback` skips the console fallback when ctx is
   already cancelled, so an already-unlocking volume doesn't flash a
   stray console prompt.

## Why

- Drops `/usr/bin/plymouth` and its transitive deps from the
  initramfs.
- Cancellation hygiene: a sibling-token-wins ctx cancel now closes
  the socket and the prompt goroutine returns immediately. The prior
  exec path left both the subprocess and the goroutine blocked
  indefinitely.
- Ctx-aware Plymouth path completes the cancellation arc started by
  #354 / #356 — keyboard, FIDO2-PIN, and TPM2-PIN prompts already
  cancel on autounlock; this extends the same to the splash prompt.

No new dependencies. Tests for splash messaging (#357) still pass
unchanged.

## Upstream Plymouth dependency

The cancel-on-hangup behavior in commit 5 — dismissing the on-screen
prompt when the booster client disconnects — depends on plymouthd
having a connection-hangup handler that tears down pending prompts.
That fix is Plymouth [MR !393](https://gitlab.freedesktop.org/plymouth/plymouth/-/merge_requests/393),
currently awaiting upstream review (closes Plymouth issues #125 and
#126).

This PR does **not** require !393 to land first. Without it, booster
still cleans up its end of the socket on cancel — boot proceeds
correctly, the goroutine returns, no orphaned subprocess. The only
visible difference is a stale prompt that lingers on the splash until
plymouth quits later in boot. Pure visual polish, not a regression vs.
the exec path.

Once !393 is in users' plymouthd, the splash prompt also dismisses
cleanly, completing the UX that #354 / #355 / #356 brought to the
console side.

## Test plan

- [x] `go build ./init ./generator` clean
- [x] `TestPromptVolumeUnlocked` and `TestTokenFriendlyName` pass
- [x] Image builds with the smaller plymouth payload (no
      `/usr/bin/plymouth` in the cpio)
- [x] Boot test on host with concurrent FIDO2 + TPM2 + keyboard
      unlock paths
- [x] Boot test on a plymouthd build without the connection-hangup
      handler — confirm no behavioural regression vs. exec path

### DIFF
--- a/init/plysocket.go
+++ b/init/plysocket.go
@@ -8,6 +8,65 @@ import (
 	"net"
 )
 
+// Direct IPC client for plymouthd's abstract Unix socket.
+//
+// Plymouth is the graphical boot splash daemon used to render password
+// prompts and progress messages during early boot:
+// https://gitlab.freedesktop.org/plymouth/plymouth
+//
+// The wire protocol verbs and reply byte constants are defined upstream in
+// src/libply-splash-core/ply-boot-protocol.h.
+//
+// Frame format for client → server commands:
+//
+//	[verb byte] [version = 2] [argLen byte] [argLen bytes, NUL-terminated]
+//
+// The server replies with a single byte for most verbs: 0x06 (ACK) on
+// success, 0x15 (NAK) on failure. The ask-password verb '*' is the
+// exception: the server replies with 0x02 (Answer) followed by
+// [uint32 length, little-endian][password bytes, not NUL-terminated].
+// plymouthAskPasswordSocket implements the answer-shaped exchange;
+// plymouthCmd handles the ACK/NAK shape used by ping, message, quit, etc.
+//
+// Verb reference (full set from ply-boot-protocol.h; only the first six are
+// exercised by booster — the rest are listed so future callers don't have
+// to cross-reference upstream):
+//
+//	IMPLEMENTED
+//	  P   PING                 plymouthPingOnce — readiness probe
+//	  $   SHOW_SPLASH          shown once plymouthd is up
+//	  *   PASSWORD             plymouthAskPasswordSocket (Answer reply)
+//	  M   SHOW_MESSAGE         plymouthMessage — status text under the prompt
+//	  Q   QUIT                 sent with retain-splash=1 at switchroot handoff
+//	  R   NEWROOT              advertises the new root path to plymouthd
+//
+//	UNUSED (defined upstream, no booster caller)
+//	  U   UPDATE               progress-bar percent
+//	  C   CHANGE_MODE          boot/shutdown/updates/firmware-upgrade/etc
+//	  u   SYSTEM_UPDATE        system-upgrade DM message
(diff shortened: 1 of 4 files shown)


init/plysocket.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	"context"
+	"encoding/binary"
+	"fmt"
+	"io"
+	"net"
+)
+
+const (
```
anatol, line 70: could you please enhance the documentation to the plymouth go file here? Links to the Plymouth upstream project and the protocol specification would be useful. Plus some short description of the protocol itself.

anatol:
Thank you very much for this change!

PR #359, opened 2026-05-07
init,generator: surface crypttab failure modes and fix cmdRoot remap

pilotstew:
## Summary

Three crypttab-adjacent failure modes that manifest as silent boot
hangs, plus the docs fix to match current behavior.

## Commits

- **generator: warn on unreadable /etc/crypttab instead of silent skip**
  93df14f conflated ENOENT (legitimately optional) with EACCES (almost
  always operator running the generator unprivileged). The latter now
  warns at build time so a broken initramfs surfaces immediately
  instead of as a hung boot.

- **init: rewrite cmdRoot to /dev/mapper/<name> on regular-loop match**
  The synthesis-fallback in `matchLuksMapping` rewrote cmdRoot to the
  post-unlock mapper path; the regular loop didn't, so adding a
  crypttab/rd.luks.* entry covering the LUKS header UUID broke any
  `root=UUID=<luks-uuid>` that previously worked via fallback.

- **init: warn when root=/dev/mapper/<name> has no LUKS unlock spec**
  Diagnostic for the silent-hang where root= names a mapper but
  nothing in cmdline or crypttab will produce it. Warning, not fatal,
  because LVM/RAID/dm-integrity also create mapper nodes outside
  luksMappings.

- **docs/manpage: correct crypttab merge behaviour (skipped → merged)**
  Wording predates 42d81fd. Crypttab options are merged into
  rd.luks.*-derived mappings, not skipped.

## Test plan

- [x] \`go test ./init/ ./generator/\` passes (sudo-only LVM/MdRaid
      tests pre-existing skip)
- [x] new unit tests: \`TestMatchLuks*\`, \`TestUnreachableMapperName\`,
      \`TestAppendCrypttabUnreadable\`
- [x] live boot: baseline (rd.luks.name= + root=/dev/mapper/cryptroot)
      boots normally
- [x] live boot: cmdRoot remap path (rd.luks.name= + root=UUID=...)
      boots normally; expect \`re-pointing root to /dev/mapper/...\`
      in dmesg
- [x] live boot: unreachable warning (root=/dev/mapper/cryptroot, no
      unlock spec) prints diagnostic before hang

## Follow-up

Discovery of these failure modes made me want to try DPS LUKS-root
partition autodiscovery as a new feature. Planning a separate PR for
that — this one is fix-only.

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -123,9 +123,14 @@ func generateInitRamfs(conf *generatorConfig) error {
 		crypttabPath = "/etc/crypttab"
 	}
 	if hasFido2, err := img.appendCrypttab(crypttabPath); err != nil {
-		if !explicitCrypttab && (os.IsNotExist(err) || os.IsPermission(err)) {
-			// default path unavailable — crypttab is optional, skip silently
-		} else {
+		switch {
+		case explicitCrypttab:
+			return err
+		case os.IsNotExist(err):
+			// optional file
+		case os.IsPermission(err):
+			warning("crypttab: %s exists but is unreadable (%v) — image will lack /etc/crypttab-derived LUKS unlock specs; re-run with sufficient privileges to include them", crypttabPath, err)
+		default:
 			return err
 		}
 	} else if hasFido2 {
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -132,7 +132,7 @@ Some parts of booster boot functionality can be modified with kernel boot parame
     Note that booster also supports LUKS v2 persistent flags stored with the partition metadata. Any command-line options are added on top of the persistent flags.
 
 Booster supports unlocking LUKS volumes declared in `/etc/crypttab` (see **crypttab(5)**).
-Only entries marked with the `x-initrd.attach` option are bundled into the initramfs at image build time. `rd.luks.*` kernel parameters take precedence — if a cmdline parameter already covers a device, its crypttab entry is skipped.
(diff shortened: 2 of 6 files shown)


generator/generator.go
```diff
@@ -123,9 +123,14 @@ func generateInitRamfs(conf *generatorConfig) error {
...
-		if !explicitCrypttab && (os.IsNotExist(err) || os.IsPermission(err)) {
-			// default path unavailable — crypttab is optional, skip silently
-		} else {
+		switch {
+		case explicitCrypttab:
+			return err
+		case os.IsNotExist(err):
+			// optional file
+		case os.IsPermission(err):
+			warning("crypttab: %s exists but is unreadable (%v) — image will lack /etc/crypttab-derived LUKS unlock specs; re-run with sufficient privileges to include them", crypttabPath, err)
```
anatol, line 132: Thank you for adding an actionable message for users. things like this make the user-side debugging much easier.

anatol:
> Discovery of these failure modes made me want to try DPS LUKS-root partition autodiscovery as a new feature. Planning a separate PR for that — this one is fix-only.

Autodiscovery would be a useful and user-friendly feature. So big +1 from me.

It just needs a good user documentation that describes how booster autodescovers and processes the partitions. This will help to reduce the future user's confusion.

I think it is a good idea to expand or restructure the man page and include a new section about autodiscovery logic.

PR #361, opened 2026-05-09
init/console: skip per-keystroke kmsg flood during password entry

### LINKED ISSUE #360: Logs flooded with asterisks from the password
Running tests with verbosity enabled I see following text in the logs

```
[    0.560174] booster: *
[    0.560327] booster: *
[    0.560469] booster: *
[    0.560637] booster: *
[    0.560800] booster: *
[    0.560940] booster: *
[    0.561077] booster: *
[    0.561206] booster: *
[    0.561337] booster: *
[    0.561467] booster: *
[    0.561596] booster: *
[    0.561750] booster: *
[    0.561890] booster: *
[    0.562041] booster: *
[    0.562181] booster: *
[    0.562320] booster: *
[    0.562461] booster: *
[    0.562599] booster: *
[    0.562761] booster: *
[    0.562902] booster: *
[    0.563050] booster: *
[    0.563190] booster: *
[    0.563326] booster: *
[    0.563465] booster: *
[    0.563604] booster: *
[    0.563748] booster: *
[    0.563884] booster: *
[    0.564023] booster: *
[    0.564159] booster: *
[    0.564293] booster: *
[    0.564426] booster: *
[    0.564559] booster: *
[    0.564717] booster: *
[    0.564846] booster: *
[    0.564976] booster: *
[    0.565110] booster: *
[    0.565239] booster:    Unlocking...
````

Having all these asterisks in the debug logs not that valuable, it should probably be skipped or replaced with just one line like '<** password typed **>'

### PR DESCRIPTION
**Pull request by pilotstew**
## Summary

Closes #360. In `-tags test` (qemu integration) builds, every `*` echo
from raw-mode password input becomes its own `/dev/kmsg` record, so a
short PIN floods the boot log with dozens of `booster: *` lines before
`Unlocking...`.

Introduce `consoleEcho` for terminal-only decoration (asterisks,
backspace, line-erase). Production behavior is unchanged — the seven
decorative call sites in `readPasswordOn` still write to stdout via
`fmt.Print`. In test builds `consoleEcho` is a no-op, replaced by one
`info("password typed (%d chars)")` breadcrumb at end of input so
test logs still confirm input occurred.

The prompt itself, status redraws, and other informative output keep
using `consolePrint`.

## Test plan

- [x] `go test ./init/` — `TestReadPassword*` (incl. PTY harness in
      `console_input_pty_test.go`), `TestKillWord`,
      `TestConsoleSetFontFailureIsNonFatal` all pass
- [x] `go vet ./init/` clean under default tags and `-tags test`
- [x] `go build ./init/` and `go build -tags test ./init/` both compile
- [x] `go test -run TestLUKS2WithName ./tests/` (real QEMU boot,
      `-tags test` init, `booster.log=debug`, 4-char passphrase):
      0 × `booster: *` lines (was 4 before the fix), exactly one
      `[    1.070599] booster: password typed (4 chars)` breadcrumb

### DIFF
--- a/init/console_input.go
+++ b/init/console_input.go
@@ -509,7 +509,20 @@ func consolePrint(msg string) {
 	}
 }
 
-// Terminal byte sequences written via consolePrint.
+// consoleEcho writes terminal-only decoration (asterisks, backspace, line-erase
+// sequences) under consoleMu. In qemu integration builds (-tags test) it is a
+// no-op: per-character writes to /dev/kmsg flood the kernel log without
+// rendering anything useful (issue #360). readPasswordOn emits one summary
+// "password typed" line at end-of-input to keep a breadcrumb in test logs.
+// Callers must hold consoleMu.
+func consoleEcho(msg string) {
+	if quirk.TestEnabled {
+		return
+	}
+	fmt.Print(msg)
+}
+
+// Terminal byte sequences written via consoleEcho.
 const (
 	eraseChar           = "\b \b"      // erase the character at the cursor
 	eraseLineAndAdvance = "\r\x1b[K\n" // wipe current line and move to next
@@ -612,7 +625,7 @@ func readPasswordOn(ctx context.Context, tty *os.File, prompt, postPrompt string
 		_ = unix.IoctlSetInt(fd, unix.TCFLSH, unix.TCIFLUSH)
 		consoleMu.Lock()
 		consolePrompt.active = false
-		consolePrint(eraseLineAndAdvance)
+		consoleEcho(eraseLineAndAdvance)
 		consoleMu.Unlock()
 		return nil, ctx.Err()
 	}
@@ -662,7 +675,7 @@ loop:
 			if !masked {
 				consoleMu.Lock()
 				consolePrompt.asterisks++
-				consolePrint("*")
+				consoleEcho("*")
 				consoleMu.Unlock()
 			}
 		case keyBackspace:
@@ -672,7 +685,7 @@ loop:
 				if !masked {
 					consoleMu.Lock()
 					consolePrompt.asterisks--
-					consolePrint(eraseChar)
+					consoleEcho(eraseChar)
 					consoleMu.Unlock()
 				}
 			}
@@ -681,12 +694,12 @@ loop:
 			if masked {
 				cp := countCodepoints(password)
 				for range cp {
-					consolePrint("*")
+					consoleEcho("*")
 				}
 				consolePrompt.asterisks = cp
 			} else {
 				for range consolePrompt.asterisks {
-					consolePrint(eraseChar)
+					consoleEcho(eraseChar)
 				}
 				consolePrompt.asterisks = 0
 			}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: init/console_input.go [hunk 1]
```diff
@@ -509,7 +509,20 @@ func consolePrint(msg string) {
...
+// sequences) under consoleMu. In qemu integration builds (-tags test) it is a
+// no-op: per-character writes to /dev/kmsg flood the kernel log without
+// rendering anything useful (issue #360). readPasswordOn emits one summary
+// "password typed" line at end-of-input to keep a breadcrumb in test logs.
+// Callers must hold consoleMu.
+func consoleEcho(msg string) {
+	if quirk.TestEnabled {
+		return
+	}
+	fmt.Print(msg)
```
anatol, line 522: How does it work with normal (non-test) case? Will the asterisk be printed to kernel logs (dmesg)?

anatol: @pilotstew i have one question here ^^

pilotstew: Apologies. I wrote this up several days ago and thought I submitted but it was in pending review status...

Short answer: no — asterisks never reach dmesg in production. The split is between `consolePrint` (kept) and the new `consoleEcho`:

```go
// consolePrint — prompt-path messages that *should* appear in test kmsg
// (prompt text, line-erase on cancel, etc.)
func consolePrint(msg string) {
    if quirk.TestEnabled {
        _, _ = fmt.Fprint(devKmsg, "<", 2, ">booster: ", msg, "\n")
    } else {
        fmt.Print(msg)
    }
}

// consoleEcho — terminal-decoration bytes only ("*", "\b \b", "\r\x1b[K\n").
// fmt.Print in production; no-op in test to avoid the per-keystroke kmsg flood.
func consoleEcho(msg string) {
    if quirk.TestEnabled {
        return
    }
    fmt.Print(msg)
}
```

**Production flow** (the normal case): `readPasswordOn` writes the prompt and any redraws through `consolePrint`, and per-keystroke asterisks/erases through `consoleEcho`. Both end up at `fmt.Print`, which writes to stdout. Init inherits stdout from `/dev/console`, so the asterisks render on whichever terminal the kernel attached as the console (vt, serial, etc.) — the same place the user is typing. `dmesg` reads the kernel printk ring, which is populated only by `/dev/kmsg` writes and in-kernel `printk()`. Neither is involved here, so `dmesg` shows nothing related to password entry — same as before this PR.

**Test flow** (`-tags test`, qemu integration): before this PR every keystroke ran through `consolePrint`, hitting `<2>booster: *` in `/dev/kmsg` — that's the flood from #360. After this PR the asterisk/erase sites use `consoleEcho`, which returns early. As a breadcrumb so test logs still record that input occurred, `readPasswordOn` emits one summary line at end-of-input:

```go
if !quirk.TestEnabled {
    console("\n")
} else if len(password) > 0 {
    // Replace the per-keystroke asterisks consoleEcho suppressed in
    // test builds with one breadcrumb line — issue #360.
    info("password typed (%d chars)", countCodepoints(password))
}
```

So in production an asterisk is a literal `"*"` byte to stdout/console. It never crosses `/dev/kmsg`, so `dmesg` never sees it.

anatol:
Thank you for the explanation. The change looks good to me.

PR #362, opened 2026-05-12
init: clear splash status synchronously at switch_root

pilotstew:
statusMessageTimed deferred its clear via a goroutine sleeping for the timeout duration. At switch_root the exec races that goroutine. This meant that in practice the exec wins and the sleep goroutine fails leaving the final "unlocked via <token>" frame visible on the splash after we hand off to systemd.

Replace with a synchronous clearSplashStatusSync() invoked at the top of switchRoot, and drop the deferred-clear from recoverTokenPassword (which was the only caller of statusMessageTimed).

As a reminder the intent is that these plymouth status messages will be removed completely once !393 makes its way to the live plymouth repo. It's interim purpose is to inform the user the prompt for the current luks partition is already unlocked via concurrent token and may be dismissed to allow for another device or the graphical splash.

### DIFF
--- a/init/luks.go
+++ b/init/luks.go
@@ -538,7 +538,7 @@ func recoverTokenPassword(ctx context.Context, volumes chan *luks.Volume, d luks
 	if !tryPassphraseAgainstSlots(ctx, volumes, d, t.Slots, password) {
 		return false
 	}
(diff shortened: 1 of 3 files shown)


anatol:
Thank you!

PR #365, opened 2026-05-18
LUKS token unlock: avoid the spurious PIN prompt when a non-interactive token can unlock

### LINKED ISSUE #255: Boot isn't silent
Booster asks for a password (`Enter passphrase for root_decrypted:`) even though I set the kernel parameter to `silent` and I unlock via TPM. Is there a way to make boot truly silent? Somehow it worked in the past but after I re-setup the boot process (because Windows messed up secure boot)  it didn't work anymore.

pilotstew:
Closes #255.

## Problem

When a LUKS2 device has both a PIN-bearing token (TPM2-PIN, FIDO2-PIN) and a non-interactive one (TPM2 PCR-only, touchless FIDO2, clevis), booster races them concurrently and draws the interactive PIN prompt the instant the PIN token dispatches — even though the non-interactive token would unlock the device silently. The reporter asked for a way to let the silent path win first.

## Changes

Two independent, opt-in mechanisms plus a related precedence fix. All are off / behaviour-preserving by default — no change unless configured.

- **`pin_delay`** (concurrent mode — what the issue literally asks for): briefly holds the *first* PIN prompt so a parallel non-interactive token can unlock first; if it does, the prompt is never rendered (cancelled before draw). If the delay expires the prompt appears as before — no unlock path is lost, the delay only postpones it.
- **`serialize_tokens`**: opts out of token concurrency entirely — tokens are tried one at a time in ascending token-ID order, so a non-interactive token enrolled before a PIN token unlocks first (behaviour closer to `systemd-cryptsetup`). Per-type bounds (`clevis_timeout` / `tpm2_timeout` / `fido2_timeout`) keep one stuck token from hanging the chain.
- **`token_timeout`** (top-level): the device keyboard-fallback timer, booster-config peer of crypttab/`rd.luks.options` `token-timeout=`, with a documented precedence order.
- **crypttab precedence fix**: an explicit kernel-cmdline `token-timeout=` was being silently overridden by a crypttab entry's *implicit* 30s default; now an explicit cmdline value wins, consistent with the keyfile/header/tries merges in `mergeCrypttabOptions`.

## Verification

Unit tests for each path (`TestPinDelay`, `TestCtxSleep`, `TestMergeCrypttabOptionsExplicitCmdlineTokenTimeoutWins`, the existing `token_timeout` suite). The `pin_delay` path is QEMU-verified end-to-end against a PIN + parallel TPM2-PCR fixture: without the option the PIN prompt is drawn (the reported symptom); with `pin_delay` long enough to outlast the racing token's unseal it is never drawn, the device unlocks silently and `switch_root` completes.

Manpage updated (CONFIG FILE bullets + the `## CRYPTTAB` precedence sentence).

### DIFF
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -81,6 +81,8 @@ Booster advantages:
 
  * `enable_fido2` is a boolean flag that enables FIDO2 hardware token support.
 
+ * `serialize_tokens` is a scoped block; set `serialize_tokens.enabled: true` (default off) to make booster try a device's LUKS tokens strictly one at a time, in ascending token-ID order, instead of racing them concurrently. Because each token is attempted to completion before the next begins, a non-interactive token (TPM2 PCR-only, touchless FIDO2, clevis) enrolled ahead of a PIN-bearing one unlocks the device before the PIN token is reached, so no PIN prompt is drawn — behaviour closer to `systemd-cryptsetup`. The trade-off is that a slow token, such as clevis waiting on the network, delays the next token instead of running alongside it. The keyboard passphrase fallback still runs last.
+
 Once you are done modifying your config file and want to regenerate booster images under `/boot` please use `/usr/lib/booster/regenerate_images`.
 It is a convenience script that performs the same type of image regeneration as if you installed `booster` with your package manager.
 
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -32,6 +32,10 @@ Booster advantages:
     vconsole: true
     enable_lvm: true
     enable_mdraid: true
+    token_timeout: 30s
+    serialize_tokens:
+      enabled: true
+      clevis_timeout: 45s
 
  * `network` node, if present, initializes the network at the boot time. It is needed if mounting a root fs requires access to the network (e.g. in case of Tang binding).
     The network can be either configured dynamically with DHCPv4 or statically within this config. In the former case `dhcp` is set to `on`.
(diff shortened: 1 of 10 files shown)


anatol:
The code itself looks good.

cc @thePrivacyFanatic and @dryBoneMarrow for more attention to this feature review.

anatol (approved):

anatol:
Thank you @pilotstew !

dryBoneMarrow:
Thank you very much, can't wait to try it out!

anatol:
If you are feeling adventurous, it would be helpful to compile Booster from HEAD and test the functionality.

Otherwise please wait a bit until there will be a new release.

dryBoneMarrow:
I am happy to test the change. I was able to compile the generator, but fido2plugin.so is missing. I wasn't able to find out how to compile it (I'm not too familiar with go and `go build` in the fido2plugin directory didn't work) and even if I was able to compile it I don't know how to point to it. If you could give me a hint on how to do this I'll go ahead and test it :)
```
(master) .../tmp.itoqyPnQD7/booster/generator> sudo ./generator build /boot/booster-linux.img --force
2026/05/21 16:35:30 fido2 plugin /usr/lib/booster/fido2plugin.so: open /usr/lib/booster/fido2plugin.so: no such file or directory
```

anatol:
@dryBoneMarrow if you are Arch Linux then the easiest way is to use `booster-git` package here https://aur.archlinux.org/packages/booster-git

Otherwise try to follow build instructions from here https://github.com/anatol/booster/blob/master/packaging/arch/PKGBUILD

dryBoneMarrow:
Thanks for the tip, I was able to test this change.
Good news: My computer isn't bricked
No so good news: boot still isn't silent, but now the displayed text is different.
I've made a video here
https://www.swisstransfer.com/d/8a151440-29d2-4dab-ad9e-59a7861f51d3
I use clevis with a tpm key (thus a non interactive token) as my main way of authentication and a password and a fido2 key both as a backup

anatol:
Thank you for sharing the video.

The messages you show are coming from two places `statusMessage(mappingName + " unlocked via " + tokenFriendlyName(t.Type))` and `statusMessage("Waiting for FIDO2 security key for " + mappingName + "...")`.

And here is the function

```go
func statusMessage(msg string) {
        if plymouthEnabled {
                plymouthMessage(msg)
        } else if msg != "" {
                consoleMu.Lock()
                if consolePrompt.active && !promptVolumeUnlocked() {
                        consolePrint("\n\n" + msg + "\n\n" + consolePrompt.text + strings.Repeat("*", consolePrompt.asterisks))
                } else {
                        consolePrint("\n\n" + msg + "\n\n")
                } 
                consoleMu.Unlock()
        }       
} 
```

It sounds like this codepath needs to take the log level into account.

pilotstew:
@anatol I added statusMessages primarily for prompt dismissal awareness. It does create issues for users wanting a truly silent boot. Would you prefer gating to when prompt is showing? An option to hide? Demote entirely to log?

anatol:
Good question!

In verbose mode, we want to print information about which method recovered the password. This helps the user understand how the password was recovered.

Quiet mode is trickier. We still have a password prompt asking the user to enter a password, and that prompt needs to be shown even when quiet mode is enabled.

For other messages, such as “waiting for FIDO USB,” the behavior is less obvious. On one hand, we want to respect the user’s choice to keep the output silent. On the other hand, if the system waits for the FIDO USB device indefinitely, the user may never understand what is happening if the message is suppressed.

pilotstew:
I inventoried/categorized all the statusMessages I've added over my PRs:

**Actionable — user must act or understand why a prompt changed**
- `"Please touch the FIDO2 key for <dev>"` (`luks.go:369`)
- `"FIDO2 skipped, falling back to passphrase"` (`luks.go:511`)
- `"FIDO2 PIN auth blocked (too many wrong attempts), falling back to passphrase"` (`luks.go:515`)
- `"FIDO2 PIN is blocked (reset required), falling back to passphrase"` (`luks.go:519`)
- `"TPM2 PIN skipped"` (`luks.go:603`)

**Prompt-dismissal context — explains why a visible passphrase prompt just vanished**
- `"<dev> unlocked via SSH"` (`ssh.go:241`)
- `"<dev> unlocked via <token>"` (`luks.go:782`)

**State announcement — purely informational**
- `"Unlocking via SSH..."` (`ssh.go:235`)
- `"Waiting for FIDO2 security key for <dev>..."` (`luks.go:424`)
- `"No FIDO2 device found for <dev>, insert security key or wait for passphrase prompt"` (`luks.go:434`)

**Clears — no-op on console, used for clearing messages on Plymouth splash**
- `""` after SSH passphrase failure (`ssh.go:245`)
- `""` after FIDO2 device acquired (`luks.go:342`, direct `plymouthMessage`)
- `""` after FIDO2 touch completes (`luks.go:386`)
- `""` after passphrase success before Plymouth quits (`luks.go:948`)

My thoughts,

1. **Actionable / Prompt-dismissal announcements** (`"... unlocked via ..."`): gate on `consolePrompt.active && !promptVolumeUnlocked()`. If no prompt is up, these are just clutter. This alone fixes the clevis-wins-silently case. 

2. **`"Waiting for FIDO2 security key ..."`**: defer behind a short timer (10-15s). If the device arrives or another token wins inside that window, the line never renders. If FIDO2 really is the slow path, the user sees it after the threshold. Addresses the silent-hang issue...
3. **`"Unlocking via SSH..."`**: leave it as is...only renders on the remote console anyway (Plymouth or under their own prompt). Would be fine to gate by kernel `quiet` too...but I kinda like the idea of leaving it. In qemu testing it was reassuring to see the message on the remote host when a SSH user was trying to unlock.
4. **`"No FIDO2 device found ..."`**: keep but gate by `quiet` (booster already parses it for Plymouth). Opting into `quiet` means opting into figuring out hangs from absent feedback, same as every other Linux init.

Does this seem right to you?

anatol:
Thank you @pilotstew for this great summary. It looks good to me.

@dryBoneMarrow @kenshaw - what do you think?

dryBoneMarrow:
Sounds great to me!

Issue #191, opened 2022-10-13
SSH remote ZFS unlocking?


sotiris-bos:
Hello, this is not an issue but a question/request.

Is there a way to enable SSH to remotely unlock an encrypted ZFS root at boot? I could not find any related documentation.

Something like this dracut module but for booster: https://github.com/gsauthof/dracut-sshd

Thanks

anatol:
Hi

Booster does not support SSH for remote unlocking. It is a large and complex protocol. Instead, booster supports Tang/EMCR protocol that is much simpler and easier (and does not expose a remote shell). See https://github.com/anatol/booster/issues/24

But the first step here would be implementing ZFS encryption support with a keyfile stored in the image. That's something I need to look at first.

The next step would be to implement handling this file as a clevis-encrypted data. 

Once it is implemented, you can easily add different locking policies for your ZFS dataset e.g.:
1. network binding - your zfs will automatically unlock only in presence of a key server in your local network
2. remote unlocking with tang (it is an equivalent of ssh unlocking you ask)
3. TPM unlocking
4. Yubikey unlocking

PR #366, opened 2026-05-20
Add SSH remote LUKS unlock

pilotstew:
Closes #320.

Picks up @kenshaw's original SSH-unlock proposal after his clearance to take it
over. Reworks against current master so the submission integrates with the
existing prompt machinery rather than walking devices itself.

## Approach

- stdlib `golang.org/x/crypto/ssh` only. No dropbear, no external daemon.
- Pubkey-only auth. `ssh_host_key` + `ssh_authorized_keys` are paths in
  `booster.yaml`, read at build time and embedded into the initramfs.
- SSH submissions land in a `pendingPrompts` registry in `init/luks.go`. The
  handler broadcasts the passphrase against every LUKS device currently
  waiting and seeds the in-boot passphrase cache so sibling volumes (e.g.
  btrfs RAID1) unlock without further prompts.
- Runs concurrently with token attempts (TPM2 PIN, FIDO2, clevis) and the
  local keyboard prompt — not gated behind `tokenWg.Wait()`. A successful
  out-of-band unlock cancels `ctx`, dismissing the in-flight
  token/keyboard prompt.

## Hardening

Pubkey-only auth, explicit `MaxAuthTries=6`, 15s handshake deadline
(slow-loris guard), 10 wrong-passphrase attempts per session before
disconnect. No PAM, no shell, no PTY, no exec, no port forwarding. Threat
model and operator guidance (host key extractability from `/boot`,
brute-force economics, dual-stack `:22` caveat) live in REMOTE UNLOCK >
Security notes in `docs/manpage.md`.

## Scope

LUKS volumes only. ZFS-native encryption (#191) uses a different unlock path
and is not covered here.

## Test plan

- `init` + `generator` unit suites pass under `-race`. New tests:
  - `TestPendingPromptsInflightFence` + `TestTrySubmitPassphraseToPendingHoldsInflight`
    — race fence preventing panic when an SSH submission races `luksOpen`'s
    watcher closing `volumes` (both demonstrably fail without the fix).
  - `TestSshPromptLoopDisconnectsAfterMaxAttempts` — per-session
    brute-force cap (demonstrably fails without the cap).
  - `TestSshReadLineSharesScanner` — pasted-input safety via shared
    console inputScanner.
  - `TestParseAuthorizedKeys*`,
    `TestReadConfigRejectsEmpty/GarbageAuthorizedKeys`,
    `TestReadConfigAcceptsLooseHostKeyPerms` — build- and parse-time
    validation.
- `tests/ssh_unlock_test.go` — 5 QEMU integration cases: single root,
  multi-device shared passphrase, unauthorized key rejection,
  wrong-then-correct retry, FIDO2-pending concurrent unlock.
- Manual QEMU: TPM2-both + Plymouth + SSH — PIN prompt dismissed cleanly,
  no redraw artifact. Paste from clipboard through real SSH client clean,
  no line-mangling.
- FIDO2/Yubikey real-hardware test deferred (fake-FIDO2 path covered; real
  Yubikey will follow on the interactive harness).

## Co-author credit

- `init/ssh: SSH server for remote LUKS unlock` preserves @kenshaw's
  co-authorship — much of the structural scaffold is from #320.
- `tests: pin SSH hostfwd to IPv4 loopback` credits @anatol — pattern
  taken from `tests-improvements` (f5e33f7).

### DIFF
--- a/init/luks.go
+++ b/init/luks.go
@@ -102,6 +102,12 @@ type promptRegistration struct {
 	d           luks.Device
 	checkSlots  []int
 	mappingName string
+	// inflight counts goroutines that submitted via trySubmitPassphraseToPending
+	// (today: SSH remote unlock) and may still be mid-UnsealVolume after
+	// senderWg.Wait() returns. luksOpen's watcher waits on this before
+	// close(volumes) so the goroutine can't send on a closed channel and
+	// panic pid 1.
+	inflight sync.WaitGroup
 }
 
 func registerPendingPrompt(p *promptRegistration) {
@@ -134,6 +140,14 @@ func trySubmitPassphraseToPending(password []byte) []string {
 	pendingPrompts.Lock()
 	snapshot := make([]*promptRegistration, 0, len(pendingPrompts.entries))
 	for p := range pendingPrompts.entries {
+		if p.ctx.Err() != nil {
+			continue
+		}
+		// Bump inflight under the same lock that owns entries — luksOpen's
+		// watcher takes this lock to remove the entry before waiting on
+		// inflight, so a snapshot-then-spawn that races the close-volumes
+		// path can't sneak in unaccounted.
+		p.inflight.Add(1)
 		snapshot = append(snapshot, p)
 	}
 	pendingPrompts.Unlock()
@@ -144,11 +158,9 @@ func trySubmitPassphraseToPending(password []byte) []string {
 		wg       sync.WaitGroup
 	)
 	for _, p := range snapshot {
-		if p.ctx.Err() != nil {
-			continue
-		}
 		wg.Add(1)
 		go func(p *promptRegistration) {
+			defer p.inflight.Done()
 			defer wg.Done()
 			if tryPassphraseAgainstSlots(p.ctx, p.volumes, p.d, p.checkSlots, password) {
 				mu.Lock()
@@ -850,19 +862,6 @@ func tryCachedPassphrases(ctx context.Context, volumes chan *luks.Volume, d luks
 }
 
 func requestKeyboardPassword(ctx context.Context, volumes chan *luks.Volume, d luks.Device, checkSlots []int, mappingName string, maxTries int) {
-	// Register this prompt so out-of-band sources (SSH remote unlock) can
-	// submit against the same device. Done before the plymouth-init wait
-	// so a remote unlock can race ahead of the splash coming up.
-	reg := &promptRegistration{
-		ctx:         ctx,
-		volumes:     volumes,
-		d:           d,
-		checkSlots:  checkSlots,
-		mappingName: mappingName,
-	}
-	registerPendingPrompt(reg)
-	defer unregisterPendingPrompt(reg)
-
 	// Wait for plymouth initialization to complete before attempting to use
 	// it. Without this, udev events can trigger LUKS password prompts while
 	// plymouthd is still starting, causing the graphical prompt to fail.
@@ -1144,6 +1143,25 @@ func luksOpen(dev string, mapping *luksMapping) error {
 		}
 	}
 
+	// Register the passphrase target for the full lifetime of luksOpen so
+	// out-of-band sources (SSH remote unlock) can submit a passphrase against
+	// the passphrase slots concurrently with token attempts — not gated behind
+	// tokenWg.Wait() like the local keyboard prompt. A successful out-of-band
+	// unlock cancels ctx, which dismisses any in-flight token / keyboard
+	// prompt via the existing ctx-cancellation path.
+	var reg *promptRegistration
+	if len(checkSlotsWithPassword) > 0 {
+		reg = &promptRegistration{
+			ctx:         ctx,
+			volumes:     volumes,
+			d:           d,
+			checkSlots:  checkSlotsWithPassword,
+			mappingName: mapping.name,
+		}
+		registerPendingPrompt(reg)
+		defer unregisterPendingPrompt(reg)
+	}
+
 	// Start keyboard/keyfile unlock after all token goroutines finish (or tokenTimeout
 	// elapses). This gives hardware tokens priority over the keyboard prompt.
 	// senderWg ensures volumes is closed if this goroutine is the last sender.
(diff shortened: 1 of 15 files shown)


init/luks.go
```diff
@@ -86,6 +86,87 @@ var passphraseCache struct {
...
+	var (
+		mu       sync.Mutex
+		unlocked []string
+		wg       sync.WaitGroup
+	)
+	for _, p := range snapshot {
+		if p.ctx.Err() != nil {
+			continue
+		}
+		wg.Add(1)
```
anatol: You can replace `wg.Add(1); go func(...)` with `wg.Go(func(...))`. Or simply run `go fix ./...` and it will fix the code for you.

init/ssh.go
```diff
@@ -0,0 +1,319 @@
...
+// submissions consume one attempt up to sshMaxPromptAttempts, after which
+// the session is disconnected.
+func sshPromptLoop(ch gossh.Channel, remote net.Addr) {
+	attempts := 0
+	for {
+		if attempts >= sshMaxPromptAttempts {
+			_, _ = io.WriteString(ch, "Too many attempts, disconnecting.\r\n")
+			return
+		}
+		_, err := io.WriteString(ch, "Enter passphrase: ")
```
anatol: How does it work if there are multiple disks that wait to be unlocked? How does the use know what disk to enter password for?

init/ssh.go
```diff
@@ -0,0 +1,319 @@
...
+//   - No echo: the console prints an asterisk per character as you type;
+//     over SSH we show nothing, so Tab (which toggles that masking on the
+//     console) does nothing here.
+//   - Ctrl+C: at the console the kernel turns Ctrl+C into a cancel signal
+//     before our code ever sees the key, so the console reader never has
+//     to handle it. SSH has no such kernel handling, so we watch for the
+//     Ctrl+C byte (0x03) ourselves and abort. The exception is pasted
+//     text: a 0x03 that arrives inside a paste is real content, not a
+//     cancel, so we leave it alone there.
+func sshReadLine(ch gossh.Channel) ([]byte, error) {
```
anatol, line 269: I see what @kenshaw meant saying that golang.org/x/crypto/ssh is too low level. Feel free to add github.com/gliderlabs/ssh if you want it, and let's hope that this package is secure enough for booster.

init/ssh.go
pilotstew: I meant to circle back to this got distracted by a race fix.... I need to cover a couple items here. Most obviously it needs to read like other unlocks do "Enter passphrase for ... " also I never finished the multidevice ssh handling. It currently disconnects after each successful unlock rather than cycling through all devices. That said the passphrase is handled by passphraseCache so devices with shared passphrases unlock automatically. I'll have a new commit coming soon

init/luks.go
pilotstew: Thank you!

pilotstew:
I should be able to get these updates done tomorrow.

pilotstew:
Almost done with testing the multidevice changes, I think I'll get this done tonight.

init/ssh.go
pilotstew: Looked at gliderlabs/ssh to make sure I'm not just being stubborn about the stdlib. There's some value in net/http-style Handler API, idle timeout config, auto-replied pty-req and would save us a few lines in the listen/dispatch path (sshRun/sshHandleConn/sshHandleSession). But it doesn't change the readline path, gliderlabs ssh.Session embeds gossh.Channel from x/crypto/ssh, so Session.Read([]byte) returns the same raw channel bytes either way. So gliderlabs doesn't provide terminal-escape stripping, bracketed-paste, line editing or Ctrl+C signal handling. Deciding whether a 0x03 byte means "cancel" or "literal paste content inside \x1b[200~...\x1b[201~" is application-level and what I addressed in the byte mode reader.

I chose to route SSH input through the same per-byte FSM as the local console prompt (init/console_input.go) so the two unlock paths behave byte-for-byte identically: same line editing, same bracketed-paste handling, same UTF-8 reassembly across reads, same ctx cancellation. inputScanner comes from the raw byte read PR (commit 571cc8c) and was built with SSH paste in mind. It allows users to paste passphrases from password managers, and modern terminal emulators bracket the pasted content with \x1b[200~ ... \x1b[201~ markers. Basic readerline would treat those markers as literal password bytes and corrupt the entry. This is I believe an improvement over ssh luks unlock systems that break with copy/paste and other handlers I built via byte mapping in consoleInput.

Swapping to gliderlabs would route the save bytes through a different wrapper without removing the per-byte loop or the 0x03 check, and would add a dependency to initramfs code that currently runs on stdlib + anatol/luks.go only.  I think you were right to want stdlib here since the savings are limited to listen/dispatch boilerplate but adds another dependency and larger security surface.

TestSshReadLineSharesScanner (init/ssh_test.go) pins the parity with the console scanner across 15 sub-cases: arrow-key strip, OSC sequence strip, BS/DEL editing, Ctrl+U kill, UTF-8 multi-byte, bracketed paste with a literal Ctrl+C inside preserved as content, etc.

init/luks.go
pilotstew: Done

init/ssh.go
pilotstew: This turned out to be a bit more work than I thought it would. It did expose a small error. When a root and nonroot device are unlocked via ssh (and really this is true globally not just for SSH...testing just exposed it here) if the root device unlocks first boot continues leaving the other device to be handled after switchroot. Not really a problem as systemd can unlock it via crypttab or the user after boot. Really this is logically correct and I don't think it needs a fix. The right answer to it is the user probably shouldn't be trying to unlock partitions in initramfs that aren't needed for boot/root. I only mention it because in testing the SSH connection disconnects before unlocking all devices listed in pendingDevices which could be confusing if someone has an unusual device config.

I'm not entirely happy with the "Enter passphrase for device1, device2" phrasing. It's a bit unclear about whats happening. Any passphrase entered is tried against all devices. As each device is unlocked the device name is removed from the reprompt. It's not intuitive but it is accurate.

pilotstew:
Let me know if anything else jumps out at you or if I need to reword the Ctrl+C comments. Otherwise I think (hope) this is ready.

anatol:
@kenshaw if you have a chance, could you please try it and let us know your opinion?

kenshaw:
Sure, I'll deploy it on some production servers this weekend and try it. Thanks.

anatol:
Okay moving forward with this change.

It would be really useful to get feedback from more users. @kenshaw if you can share your experience with this implementation it would be fantastic. Thank you!

PR #367, opened 2026-05-25
init/luks: ctx-aware FIDO2 unlock primitives

pilotstew:
## Context

Foundation for the broader FIDO2 reliability work — hidraw add-event broadcasting and pre-flight token gating both fan out multiple goroutines that race on hidraw enumeration and user interaction. Those goroutines need clean cancel semantics so a winner can tear down losers immediately; without that, leaks (this PR) and contention bugs become structural rather than incidental.

## Problem

Three blocking calls in the FIDO2 unlock path do not observe `ctx.Done()`, so a goroutine can pin past its useful lifetime — when a sibling token wins, when keyboard fallback unlocks the volume, or when init exits:

| Site | Primitive |
|---|---|
| `init/luks.go:204` (`recoverFido2Password`) | `usbhidWg.Wait()` |
| `init/luks.go:304` (`recoverSystemdFido2Password`) | `usbhidWg.Wait()` |
| `init/luks.go:216` (`recoverFido2Password`) | `fido2Mu.Lock()` |

Concrete leak paths:

- A boot with no USB HID device ever binding (no internal keyboard, no FIDO key) blocks `usbhidWg.Wait()` until `switch_root`.
- Multi-FIDO2-token boots where one token unlocks first or where keyboard fallback wins leave sibling goroutines parked on `fido2Mu.Lock()` — they can't observe `cancel()` from the winning path.

## Changes

Each blocking primitive is replaced with a channel-based equivalent and a ctx-aware helper in the style of the existing `ctxSleep`:

- **`usbhidReady chan struct{}`** replaces `usbhidWg sync.WaitGroup`. The existing `usbhid sync.Once` closes the channel exactly once on first usbhid bind event. `waitForUsbhid(ctx)` selects on `<-usbhidReady` vs `<-ctx.Done()`.
- **`fido2Sem chan struct{}` (1-slot)** replaces `fido2Mu sync.Mutex`. `acquireFido2Lock(ctx)` selects on `fido2Sem<-` vs `<-ctx.Done()`. `releaseFido2Lock()` drains the slot.

The three call sites return `nil, ctx.Err()` on cancel and emit an `info()` log line.

## Verification

Unit tests (`init/luks_ctx_test.go`):
- `TestWaitForUsbhidRespectsCtxCancel` / `TestAcquireFido2LockRespectsCtxCancel` — cancel returns `ctx.Err()`.
- `TestWaitForUsbhidReturnsOnSignal` / `TestAcquireFido2LockReturnsOnRelease` — signal/release unblocks normally.

VM tests (`tests/fido2_test.go`):
- `TestCtxAwareFido2CancelOnFallback` — boots `systemd-fido2-nodev.img` with `token-timeout=5`, types fallback passphrase, asserts the cancel-path log fires before `Hello, booster!`. Without the fix, the FIDO2 goroutine pins on `sync.WaitGroup.Wait()` and the log can never fire — this is the regression gate.
- `TestCtxAwareFido2CancelMultiToken` — builds a 3-token image at test time (sudo-free, `cryptsetup token import` against the regular file), asserts the cancel log fires three times under multi-token contention.


anatol:
Thank you!

PR #368, opened 2026-05-25
init/luks: broadcast hidraw events + touch-anytime FIDO2 retry

pilotstew:
## Context

Companion to #367 (ctx-aware FIDO2 primitives). Both PRs are independent foundation work for reliable multi-token FIDO2 boots: #367 makes blocking sync primitives observe `ctx.Done()`; this PR fixes event delivery and adds touch-anytime UX.

### 1. Shared hidraw channel starves siblings

`hidrawDevices` is a single global channel. The first goroutine to read wins the device-add event; siblings starve.

- **Same disk, primary + backup key** (real-world setup) — N FIDO2 tokens enrolled on the same volume, one per physical key. The goroutine whose credential matches the inserted key never sees the device-add event if a sibling reads it first, and times out even though the right key is physically present.
- **Same key, multiple tokens** (dev/test setup — e.g. one PIN-required + one touchless on the same physical key) — only the first goroutine ever sees the device; the second never blinks. Not a typical user configuration, but useful for exercising the multi-token dispatch path without two physical keys.

### 2. Touch-only retry costs a re-prompt

A touchless FIDO2 device that returns a non-fatal error (typically "user not present" — the user didn't touch in time) gives up immediately and falls through to the next attempt or keyboard fallback. The user has to re-trigger the boot sequence to try again.

### Per-listener broadcast (commit 1)

Replace the global channel with a registry of per-call listener channels. Each `recoverSystemdFido2Password` call registers its own buffered channel via `registerHidrawListener()`; `broadcastHidrawDevice()` delivers each udev event to every listener non-blocking (a full listener buffer drops the event for that listener only).

A `done <-chan struct{}` parameter on the device-iteration loop ensures the deferred `dropListener()` actually runs on early exit.

### Touch-anytime re-seed (commit 2)

`reseedTouchlessFido2` re-injects the device into the goroutine's own listener channel on non-fatal errors, so the loop keeps retrying. The user can then touch the token at any point during boot — including while the keyboard passphrase prompt is showing — and the unlock succeeds.

Three gates:
- **!pinRequired**: PIN tokens have already engaged the user via prompt; silent re-seed would confuse the prompt state.
- **elapsed > 1s**: bounds hot-loops on fast-fail libfido2 errors that return in microseconds.
- **ctx not done**: sibling-token win cancels parent ctx; helper returns `ctx.Err()` rather than re-seeding into a doomed channel. The select on `ctx.Done()` during send avoids deadlock if the listener buffer is full when the context cancels.

Wrong-credential errors are filtered earlier by the existing skip path, so re-seed only runs on genuine "user not present" outcomes.

## Verification

`init/hidraw_broadcast_test.go` (7 tests):
- Broadcast delivers to all listeners; doesn't block on slow listener; ignores dropped listener; multi-event ordering; empty-registry no-op.
- Listener self-send (used by re-seed) reaches only the sending goroutine, not siblings; self-send and broadcast queue independently.

`init/fido2_reseed_test.go` (6 tests):
- Happy path re-seeds; PIN gate skips silently; fast-fail gate skips silently; gate boundary (elapsed==1s) is no-op (strict `>`).
- Pre-cancelled ctx returns `ctx.Err()` without sending or mutating `seen`.
- Full-buffer listener + mid-send cancel: select unblocks via `ctx.Done()` so the goroutine doesn't deadlock holding `fido2Mu`.

### DIFF
--- a/init/fido2_reseed_test.go
+++ b/init/fido2_reseed_test.go
@@ -1,0 +1,154 @@
+package main
+
+// Tests for reseedTouchlessFido2 — the gate-and-send helper that implements
+// touch-anytime FIDO2 retry (project plan item N).
+//
+// When a touchless FIDO2 device returns a non-fatal error, the goroutine
+// re-seeds the device back into its own listener channel so the recover
+// loop tries again. The user can then touch the token at any point during
+// boot — including while a passphrase prompt is showing — and the unlock
+// succeeds without further intervention.
+//
+// Three gates protect against misuse:
+//   1. !pinRequired   — PIN tokens have engaged the user via prompt; silent
+//                        re-seed would be confusing.
+//   2. elapsed > 1s   — bounds hot-loops on fast-fail libfido2 errors.
+//   3. ctx not done   — sibling-token win cancels parent ctx; don't keep
+//                        re-seeding into a doomed channel.
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestReseedHappyPath(t *testing.T) {
+	// Touchless device, slow assertion (>1s elapsed), live ctx → re-seeds:
+	// listener receives devName and seen[devName] is cleared.
+	ctx := context.Background()
+	listener := make(chan string, 1)
+	seen := set{"hidraw0": true}
+
+	err := reseedTouchlessFido2(ctx, listener, "hidraw0", seen, false, 2*time.Second)
+	require.NoError(t, err)
+
+	select {
+	case got := <-listener:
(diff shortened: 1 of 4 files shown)


anatol:
@pilotstew there is a code conflict.

Thank you for looking at it. The concurrency handling with FIDO/device initialization is one of the trickiest issues with the project currently. Thank you for taking care of it!

pilotstew:
So true, I've been working on a better solution here. It's close to done.

pilotstew:
Rebased.

anatol:
Fantastic! Thank you @pilotstew

PR #371, opened 2026-05-27
init/luks: pre-flight FIDO2 credentials before prompting for PIN

pilotstew:
## Background

The recent FIDO2 plumbing work — #367 (ctx-aware unlock primitives) and #368 (per-listener hidraw broadcast + touch-anytime re-seed) — covered the infrastructure: how FIDO2 goroutines cancel on ctx, how multiple unlock goroutines observe device-add events independently, and how touchless devices retry across boot. They did not address **which device receives the PIN prompt** when multiple tokens are enrolled.

## Problem

Within each enrolled FIDO2 token's `recoverSystemdFido2Password`, the loop walks every visible hidraw and calls `recoverFido2Password` against each. That inner function calls `askPasswordWithFallback` for the PIN before it knows whether the device holds **this token's** credential. The assertion is then sent to the device, which rejects with `FIDO_ERR_NO_CREDENTIALS` — the user has already typed a PIN against a device that could never validate it.

Example: a LUKS volume enrolled with two FIDO2 keys (older USB-A, newer USB-C Nano), only the Nano inserted at boot. The USB-A token's enrollment is encountered first; booster prompts for the USB-A key's PIN; submission is rejected by the Nano because the credential doesn't match.

## Solution

Pre-flight credential presence on each FIDO2-capable hidraw **before** prompting for the PIN. The pre-flight is a CTAP2 assertion with `up=false`, no PIN, and no HMAC-secret extension — the device returns `FIDO_OK` when the credential is present and `FIDO_ERR_NO_CREDENTIALS` when it isn't. Per CTAP 2.1 §6.2.4 neither outcome decrements the PIN-retry counter and no touch is required. Non-matching devices are skipped silently; the serial dispatcher advances to the next FIDO2 token (or, when all tokens exhaust, the keyboard prompt).

Tokens enrolled with `fido2-uv-required=true` are exempt per CTAP 2.1 §7.4 (some authenticators reject `up=false` when UV is required); those tokens proceed directly to the full assertion. Mirrors systemd's `libfido2-util.c` handling.

## Commits

1. `init/fido2iface,fido2plugin: add Fido2Preflight credential pre-check` — interface method + libfido2 implementation (combined so each commit individually builds)
2. `init: add fido2Preflight shim (cgo + nocgo)` — wires the plugin call into init; nocgo stub for non-FIDO2 builds
3. `init/luks: add hidrawSysPath test seam and failing pre-flight test` — `hidrawSysPath` becomes a package var so tests can override; introduces the failing test that the next commit makes green
4. `init/luks: gate FIDO2 PIN prompt behind pre-flight credential check` — main change: restructures the per-device loop to call preflight before the assertion
5. `init/luks: don't give up on FIDO2 before any key is plugged in` — `len(dir) > 0` guard preserves the empty-hidraw "wait for udev" semantic that the new bail-out would otherwise break
6. `init/luks: tests for the pre-flight gate` — 5 unit tests covering green-light, multi-token serial dispatch advance, empty-hidraw + late hot-plug paths
7. `docs/manpage: document FIDO2 credential pre-flight` — paragraph in the LUKS unlock concurrency section

## Testing

- ``go build`` clean for both cgo and nocgo build paths
- ``go vet ./...`` clean
- Each commit individually builds and vets cleanly
- 5 new unit tests pass under ``-race``
- End-to-end QEMU verification across ``passphrase``, ``tpm2``, ``tpm2-pin``, ``tpm2-both``, ``fido2-nodev`` scenarios — no regressions in any unlock path


init/luks_preflight_test.go
```diff
@@ -0,0 +1,410 @@
...
+	withFakeHidrawDevices(t, []string{"hidraw0", "hidraw1"}, func() {
+		token := luks.Token{
+			Type:    "systemd-fido2",
+			ID:      3,
+			Slots:   []int{4},
+			Payload: makeFido2TokenPayload(t, credID, []byte("salt"), "io.systemd.cryptsetup", false, true, false),
+		}
+		_, err := recoverSystemdFido2Password(context.Background(), token, "cryptroot")
+		if err != nil {
+			t.Fatalf("expected unlock to proceed, got error: %v", err)
```
anatol: Many of these `t.Fatalf` can be replaced with `https://github.com/stretchr/testify` equivalents like `require.NoError(t, err, "expected unlock to proceed")`

pilotstew: Done — pushed in 2d2e44b (autosquashed). 22 of 24 `t.Fatalf` calls converted to `require.X` equivalents; the remaining two are `case <-time.After(...): t.Fatal(...)` select-timeout cases where the receiver pattern reads cleaner without testify. Also folded the polling helper into `require.Eventually` and dropped the now-unused `errors` import.

anatol: `t.Fatal(...)` could probably be replaced with `require.Fail()` for consistency. But it is minor cleanup.

anatol (approved):

init/luks_preflight_test.go
pilotstew: Thanks!

anatol:
Fantastic, thank you!

PR #372, opened 2026-05-28
init/luks: eager FIDO2 PIN prompt + defer pinless "no device" hint

pilotstew:
## Background

Previously we waited for a FIDO2 device to be present before prompting for PIN entry. That meant a PIN-required token with no device could hang the FIDO2 goroutine until `tokenTimeout`, blocking a pinned TPM (or any subsequent token) from ever reaching its own prompt. The status messages "Waiting for FIDO2 security key for X..." and "No FIDO2 device found..." informed the user but didn't unblock the dispatcher.

The LUKS token JSON already tells us whether a PIN is required (`fido2-clientPin-required`), so we don't need a device in hand to know we'll need a PIN. Better to prompt eagerly from header metadata, then check for a device on submission and return a skip-sentinel if it's missing — the dispatcher advances to the next token instead of parking.

For pinless tokens there's no eager prompt to fire, but we don't want the "No FIDO2 device found" hint flickering on every boot where a faster non-interactive token (TPM2, clevis) was about to win. Defer it behind `tokenTimeout/2` (capped at 10s) — slow tokens get time to unlock first, and the hint only surfaces when it's actually useful.

Relatively free fix; should just work.

## Commits

1. `init/luks: eager FIDO2 PIN prompt for PIN-required tokens` — new `recoverFido2WithEagerPrompt` driving the prompt-first flow + `errFido2Skipped` sentinel for empty-Enter advance. 5 unit tests cover empty-Enter, reprompt-on-missing-device, PIN-attempt cap, happy path, and a regression guard for the next commit's behavior.
2. `init/luks: defer pinless "No FIDO2 device found" behind tokenTimeout/2 timer` — replaces the upstream's immediate-fire status message with a per-mapping timer (dedup via `sync.Once`).
3. `init/luks: trim 'to passphrase' from FIDO2 PIN prompt` — wording fix; empty-Enter actually advances to the next token (TPM2-PIN, more FIDO2, etc.), only reaching passphrase if all tokens exhaust.
4. `init/luks: drop waitForUsbhid block from pinless FIDO2 path` — both pinless and eager paths previously blocked on `waitForUsbhid` before the timer/rescan could fire. On usbhid-less configs (QEMU without USB, headless servers, PS2/i2c-only HID) `usbhidReady` never closes; `registerHidrawListener` already catches late `add` events independently, so the wait was only delaying the user.

## Testing

- `go build` and `go vet ./...` clean
- Each commit individually builds cleanly
- 5 new unit tests pass — testify throughout
- Plymouth-graphical QEMU verification on `fido2-nodev-pin`: Flow A (empty-Enter → fallback at t≈4s vs t≈32s before), Flow B (typed-PIN, no device → "FIDO2 device not detected — Enter FIDO2 PIN…" reprompt → empty-Enter → fallback → unlock)
- Smoke matrix on the rebased branch: `passphrase`, `tpm2`, `tpm2-pin`, `fido2-nodev`, `fido2-nodev-pin`, `clevis-tang` — no regressions

### DIFF
--- a/init/luks.go
+++ b/init/luks.go
@@ -545,15 +545,15 @@ func recoverSystemdFido2Password(ctx context.Context, t luks.Token, mappingName 
 
 	statusMessage("Waiting for FIDO2 security key for " + mappingName + "...")
 
-	// Register BEFORE waiting on usbhidReady so any 'add' events arriving
-	// after the wait but before /sys/class/hidraw/ is read are buffered for us.
+	// Register BEFORE reading /sys/class/hidraw so any 'add' events arriving
+	// during/after the scan are buffered. We deliberately do NOT block on
+	// waitForUsbhid here: on usbhid-less configs (QEMU without USB,
+	// embedded systems, laptops with only PS2/i2c HID) usbhidReady never
+	// closes, and blocking would prevent the deferred "No FIDO2 device
+	// found" timer below from ever arming. A late-arriving usbhid bind
+	// surfaces its hidraw through the listener channel.
 	hidrawDevices, dropListener := registerHidrawListener()
 	defer dropListener()
-
-	if err := waitForUsbhid(ctx); err != nil {
-		info("FIDO2 unlock for %s cancelled before USB HID ready: %v", mappingName, err)
-		return nil, err
-	}
 
 	dir, err := os.ReadDir(hidrawSysPath)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


anatol (approved):

pilotstew:
Thanks @anatol once this merges I have the final PR in this series...the promised status messaging fixes. Also do you want a PR for the clevis.go bump?

anatol:
> for the clevis.go bump

thank you for the reminder. I just bumped the clevis.go dependency. PTAL.

PR #373, opened 2026-05-28
init/luks: pre-flight eager FIDO2 PIN prompt against connected devices

pilotstew:
In my haste to get #372 out last night I missed a couple of live-test
TODOs from my notes: unplug/late-plug between primary and backup
Yubikeys. Running through them exposed a logic error in the eager-
prompt path's pre-flight.

Symptom: backup Yubikey plugged in, primary not — the prompt still
fires for the primary token and reprompts "FIDO2 device not detected"
for any PIN entry until empty-Enter dismisses it. Only then does the
backup token's prompt fire.

Fix: move pre-flight to the top of the loop. No connected hidraw
holding the credential → return `errFido2FallbackToKeyboard`
immediately, prompt never fires, dispatcher advances to the next
token. The "device not detected" reprompt path goes away.

Tests: new `TestEagerPromptSkipsWhenNoMatchingDevice`; existing
empty-Enter / fires-before-usbhid tests now put a device in place so
they exercise the eager path past the gate; the now-obsolete reprompt
test is deleted. Live-tested on the actual two-Yubikey boot.

### DIFF
--- a/init/luks.go
+++ b/init/luks.go
@@ -772,7 +772,7 @@ func recoverFido2WithEagerPrompt(ctx context.Context, mappingName string, credID
 	}
 
 	// No waitForUsbhid: findMatchingFido2Device below rescans /sys/class/hidraw
-	// on every PIN submission, so a late-arriving key is picked up next round.
+	// on every iteration, so a late-arriving key is picked up next round.
 
 	if err := acquireFido2Lock(ctx); err != nil {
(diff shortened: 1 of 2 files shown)


pilotstew:
My apologies @anatol I should have caught this before #372

anatol:
Thank you!

PR #374, opened 2026-05-28
init: console-mode prompt and log output cleanup

pilotstew:
This PR descends from a review-thread inventory in #365 where I
classified every `statusMessage` I had added across the FIDO2 / TPM2 /
clevis paths and proposed how to treat each: gate prompt-dismissal
context on an active prompt, defer "Waiting for FIDO2" behind a short
timer, keep "Unlocking via SSH..." (remote-only audience), keep "No
FIDO2 device found" gated by `quiet`.

Implementation drifted from that plan in two ways worth flagging:

1. **"Waiting for FIDO2"** ended up dropped entirely (not deferred).
   The eager-prompt path in #372 takes over for PIN-required tokens —
   the PIN prompt itself signals "we're waiting" — and the deferred-
   timer idea got applied to "No FIDO2 device found" instead. Same
   "only fire if something is actually wrong" goal, better placed.

2. **"FIDO2 skipped" and "TPM2 PIN skipped"** were classified as
   actionable in the inventory but ended up dropped. With the eager-
   prompt path landed, empty-Enter advances the dispatcher immediately
   and the next prompt (TPM2-PIN or passphrase) appears within
   milliseconds — that prompt IS the notification. A separate
   "X skipped" line just adds noise to the same UI event.

Scope also grew to include two console-output discipline fixes
(commits 3 and 4) that surfaced during the QEMU console-mode
verification matrix for #372 — log lines clobbering active prompts,
and "Unlocking..." rendering on the same line as typed asterisks.
Same "don't trample a visible prompt" theme so they ride along here.

"X unlocked via Y" was planned for gating; the gate is in place but I
had hoped [plymouth!393](https://gitlab.freedesktop.org/plymouth/plymouth/-/merge_requests/393)
would land before v.13 to obsolete the message entirely (Plymouth
auto-dismisses prompts when the client disconnects). !393 hasn't
gotten any review attention upstream, so the gate stays for now.

## Commits

1. **drop premature "Waiting for FIDO2" status** — fired unconditionally
   as the first action of `recoverSystemdFido2Password`, even when the
   key was already plugged in. "No FIDO2 device found" and "Please
   touch the FIDO2 key" already cover the real UI states.

2. **prompt-gate fallback messages, drop redundant ones** — new
   `statusMessageIfPrompt` helper (shows under Plymouth or when a
   console prompt is active). Applied to the legacy FIDO2 path +
   `recoverSystemdTPM2Password`: drop "FIDO2 skipped" / "TPM2 PIN
   skipped" entirely (next prompt is the notification); gate "PIN auth
   blocked" / "PIN is blocked" to active-prompt only; gate "X unlocked
   via Y" with the TODO above.

3. **route console log prints through prompt-redraw helper** —
   `info`/`warning`/`severe` previously printed via `fmt.Println`
   without taking `consoleMu` or re-painting the prompt. New
   `consolePrintWithPromptRedraw` acquires the mutex, writes the
   message, re-paints the active prompt + typed asterisks below so the
   cursor stays at the bottom.

4. **render password postPrompt on its own line** — the "Unlocking..."
   text rendered on the same line as the typed asterisks
   (`Enter passphrase: ********   Unlocking...`). Prepend a newline so
   it sits below the prompt.

## Testing

- `go build` and `go vet ./...` clean; each commit builds cleanly
- FIDO2 / preflight / eager-prompt / console suite passes
- QEMU-verified in a 16-scenario matrix (Plymouth graphical + console
  graphical) earlier this week

### DIFF
--- a/init/console_input.go
+++ b/init/console_input.go
@@ -89,6 +89,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 	"sync"
 	"unicode/utf8"
 
@@ -506,6 +507,20 @@ func consolePrint(msg string) {
 		_, _ = fmt.Fprint(devKmsg, "<", 2, ">booster: ", msg, "\n")
 	} else {
 		fmt.Print(msg)
+	}
+}
+
+// consolePrintWithPromptRedraw writes msg to the console under consoleMu,
+// re-painting any active password prompt below so the user's cursor stays
+// at the bottom. Used by log levels (info, warning, severe) which would
+// otherwise overwrite the prompt line mid-input.
+func consolePrintWithPromptRedraw(msg string) {
+	consoleMu.Lock()
+	defer consoleMu.Unlock()
+	if consolePrompt.active && !promptVolumeUnlocked() {
+		consolePrint(msg + "\n" + consolePrompt.text + strings.Repeat("*", consolePrompt.asterisks))
+	} else {
+		consolePrint(msg + "\n")
 	}
 }
 
--- a/init/console_input.go
+++ b/init/console_input.go
@@ -752,7 +752,9 @@ loop:
 	}
 
 	if postPrompt != "" {
-		console(postPrompt)
+		// Newline before the postPrompt so it renders on its own line below
+		// the typed asterisks, not glued to the right of them.
+		console("\n" + postPrompt)
 	}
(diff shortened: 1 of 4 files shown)


anatol:
Looks good, thank you!

PR #377, opened 2026-06-05
init/plymouth: fix DRM device initialization for Plymouth 22.x

basploeger:
Plymouth (version 22.02.122, latest version provided by Void Linux) uses two successive checks before accepting a DRM device:
1. `udev_device_get_is_initialized()` — requires any db file to exist (I:1)
2. `udev_device_has_tag(device, "seat")` — requires a `G:seat` line in the db

The previous implementation only created db files for card devices (card0,
card1) and omitted the `G:seat` and `G:uaccess` tags, so Plymouth failed
check 2 and fell back to the text (details) plugin after an 8-second timeout.

This affects systems that use an older version of Plymouth with booster and have a native DRM
driver (i.e. not simple-framebuffer / simpledrm), since booster runs without
udevd and those db files would otherwise never be created.

This commit:
- Covers all three DRM entry types: card nodes, render nodes (renderD128…),
  and connector/output entries (card0-eDP-1, card0-HDMI-A-1, …)
- Adds `G:seat` and `G:uaccess` to device entries as needed,
  matching what a real udevd would write via 70-seat.rules / 73-seat-late.rules
- Uses the correct db key format for connector entries (`+drm:<sysname>`)
  since they have no `dev` file

### DIFF
--- a/init/plymouth.go
+++ b/init/plymouth.go
@@ -347,45 +347,57 @@ func isSimpledrm(devPath string) bool {
 	return filepath.Base(driverLink) == "simple-framebuffer"
 }
 
-// createDrmUdevEntries creates udev database files for DRM card devices.
-// Plymouth uses udev_device_get_is_initialized() to check whether a DRM
-// device is ready. This function checks for /run/udev/data/c<major>:<minor>
-// files. Without udevd running, these files don't exist and Plymouth skips
-// every DRM device, falling back to the details plugin (text-based fallback) after an 8s timeout.
+// createDrmUdevEntries creates udev database files for all DRM devices.
+// Without udevd running, the database files under /run/udev/data/ don't exist
+// and Plymouth fails two successive checks for every device:
 //
-// This follows the same pattern as devMapperUpdateUdevDb() in udev.go.
+//  1. udev_device_get_is_initialized() — requires any db file to exist (I:1).
+//  2. udev_device_has_tag(device, "seat") — requires a G:seat line in the db.
+//
+// Both checks must pass; if either fails Plymouth skips the device. With no
+// usable DRM device Plymouth falls back to the text (details) plugin.
+//
+// Three kinds of entries are needed:
+//   - Card devices (card0, card1): character devices → /run/udev/data/c<major>:<minor>
+//   - Render nodes (renderD128, ...): character devices → /run/udev/data/c<major>:<minor>
+//   - Connector/output entries (card0-eDP-1, card0-HDMI-A-1, ...): no dev
+//     file → /run/udev/data/+drm:<sysname>
+//
+// All three types need G:seat; card/render nodes also get G:uaccess (matching
+// what a real udevd would write via the 70-seat.rules / 73-seat-late.rules).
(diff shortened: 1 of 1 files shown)


anatol:
cc @pilotstew for a feedback

pilotstew:
Interesting, on current Plymouth, device acceptance is just `is_initialized` + a devnode, no tag check. I couldn't find anywhere current Plymouth reads `G:seat`. So I had AI do some digging...turns out that check lived in the old 0.9.x series (`create_seats_for_subsystem` — the "non-graphical seat" regression, [Debian #751726](https://bugs.debian.org/751726)) and was dropped upstream by 24.x. For the Debian-based distros `Ubuntu 22.04 (0.9.5), Pop 22.04 / Mint 21 (jammy's 0.9.5), and Debian 11/12 (0.9.5 / 22.02)` still on pre-24 Plymouth all still have the live `udev_device_has_tag(device,"seat")`, so `G:seat` on the card is needed for compatibility.

I also had AI look at what real udevd writes (systemd 247→260), and yours mostly match: the card *and* all the connectors get `G:seat` (+`master-of-seat`) — the `card[0-9]*` glob in `71-seat.rules` catches the connector entries too...so tagging the connectors is correct.

I couldn't verify the **render node** (`renderD128`). As far as I can tell real udev never puts `seat` on it (only `card[0-9]*` gets seat, which `renderD` doesn't match; it gets `uaccess` only, and only on the distros that build with `group-render-mode=0660`). On current Plymouth your `G:seat` there is harmless...version 24.x explicitly ignores render nodes but on the older seat-checking builds it actually makes Plymouth *open* the render node and fail (`drmModeGetResources`→NULL, "Could not get card resources") before falling through to the card. Real udev wouldn't trigger that, since without a seat tag Plymouth 24.x skips it. Is this something you ran into on your system? If it's not solving a real problem removing `G:seat` from render nodes allowing plymouth to skip is cleaner, even if it fails gracefully on `drmModeGetResouces`. This could be a distro / system specific necessity and looks safe either way.

All that is to say I think the changes generally look good, they allow for distro specific backwards compatibility. My only nit is that for future reference the commit message isn't quite accurate given the above. But tbh it's hard for me fully evaluate the differences in udev/plymouth handling across distros/versions.

FWIW I regression-tested the change against current Plymouth (24.004.60, booster in QEMU/virtio-gpu): graphical splash, prompt, and unlock all work end-to-end, no fallback to details — so no regression on 24.x.

What distro / plymouth version is the affected system?

basploeger:
@pilotstew Thanks for the thorough review! 

The code changes were made by Claude AI and made Plymouth v 22.02.122 work on my Void Linux laptop (Lenovo Thinkpad T460s). Looks like the package is outdated on Void, but that's what we've got there.

I made a change to not add `G:seat` to render devices and tested that it still works on my system, so should be safe to remove that tag.

I have updated the PR description as well. Let me know what you think.

pilotstew:
Great work @basploeger! Thank you for discovering it and patching it!

anatol:
Thank you @basploeger for your work!

PR #378, opened 2026-06-07
generator: Automatically enable LVM if unspecified and we detect it above the rootfs

NHellFire:
This will look through /sys to determine the types of the parent block devices and then enable LVM if we're a child of LVM and enable_lvm hasn't explicitly been disabled (similar to what dracut does)

I've managed to do this with only adding one tiny library requirement to avoid having to parse /proc/self/mountinfo myself. I do still need to add it to the end-to-end tests, but I'm posting it now for feedback.

This can probably also easily be extended to do the same for enable_mdraid

### DIFF
--- a/generator/rootmountinfo.go
+++ b/generator/rootmountinfo.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/anatol/booster/generator/quirk"
 	"github.com/moby/sys/mountinfo"
 )
 
@@ -31,7 +32,14 @@ func (o *RootMountInfo) GetSysfsPath(majmin string) string {
 }
 
 func (o *RootMountInfo) GetRootDevice() error {
-	mnts, err := mountinfo.GetMounts(mountinfo.SingleEntryFilter("/"))
+	mntPoint := "/"
+	if quirk.TestEnabled {
+		if tmpMnt := os.Getenv("BOOSTER_TEST_ROOT_MOUNTPOINT"); tmpMnt != "" {
+			mntPoint = tmpMnt
+		}
+	}
+
+	mnts, err := mountinfo.GetMounts(mountinfo.SingleEntryFilter(mntPoint))
 	if err != nil {
 		return err
 	} else if len(mnts) == 0 {
--- a/generator/rootmountinfo.go
+++ b/generator/rootmountinfo.go
@@ -44,13 +44,13 @@ func (o *RootMountInfo) GetRootDevice() (error) {
 }
 
 func (o *RootMountInfo) GetBlockDeviceInfo(devPath string) (*BlockDeviceInfo, error) {
-	fpDevNum, err := os.Open(devPath + "/dev")
+	devNum, err := os.ReadFile(devPath + "/dev")
 	if err != nil {
 		return nil, err
 	}
 
 	var devMajor, devMinor uint32
-	num, err := fmt.Fscanf(fpDevNum, "%d:%d", &devMajor, &devMinor)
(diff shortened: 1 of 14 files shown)


anatol:
Could you please also add an integration test for the feature?

cc @pilotstew

generator/rootmountinfo.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (o *RootMountInfo) GetBlockDeviceInfo(devPath string) (*BlockDeviceInfo, error) {
+	fpDevNum, err := os.Open(devPath + "/dev")
+	if err != nil {
+		return nil, err
+	}
+
+	var devMajor, devMinor uint32
+	num, err := fmt.Fscanf(fpDevNum, "%d:%d", &devMajor, &devMinor)
```
pilotstew: os.Open is never closed. ReadFile might be a better option here:

````
        data, err := os.ReadFile(devPath + "/dev")
        if err != nil {
                return nil, err
        }

        var devMajor, devMinor uint32
        num, err := fmt.Sscanf(string(data), "%d:%d", &devMajor, &devMinor)

**Comment by pilotstew** [comment]
Great feature, I like the recursive /sys walk, it handles the stacked cases well.  Could use gofmt.
  
   Auto-enable shares the hard-fail path with explicit enable_lvm: true but missing lvm binary, small risk here with auto enable....once working build inexplicably fails.  Minor.
  
  If you decide to take on the mdraid extension, RAID/integrity LV roots will detect as "needs LVM" but the bundled module set (dm_mod/snapshot/mirror/cache/thin) doesn't include dm_raid/dm_integrity, so they won't assemble yet.

File: generator/rootmountinfo.go (hunk 1 above)
Comment (NHellFire) [reply generator/rootmountinfo.go]: Thanks, I missed that

**Comment by NHellFire** [comment]
> Could you please also add an integration test for the feature?

I'll hopefully have time to look at that tomorrow.

**Comment by NHellFire** [comment]
Tests added now. To avoid having to make more changes (such as pre/post scripts in buildVmInstance) , I went with just generating an initramfs against lvm.img and checking the generated init config.

**Comment by anatol** [comment]
Thank you for your work. The feature looks good. But the test needs changes. The e2e test use cached assets as a way to reduce test time. And tests should not modify someone's else assets - if the assets are distinct enough then it should be it's own asset/image. It is ok to have a common generator script if the generation logic is close enough.

PS I have a change sitting in the queue to make the assets completely readonly https://github.com/anatol/booster/commit/843bf54bfb4e90bc6eb73ab85289e59a9c38c384 thus avoiding any corruption issues with crashed tests.

**Comment by NHellFire** [comment]
I've changed it to generate its own assets, as well as rebased. Sorry for the delay on this!

File: generator/rootmountinfo.go [hunk 2]
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (o *RootMountInfo) GetSysfsPath(majmin string) string {
+	return "/sys/dev/block/" + majmin
+}
+
+func (o *RootMountInfo) GetRootDevice() error {
+	mntPoint := "/"
+	if quirk.TestEnabled {
+		if tmpMnt := os.Getenv("BOOSTER_TEST_ROOT_MOUNTPOINT"); tmpMnt != "" {
```
anatol, line 37: I would prefer not to use envvars here.

can the rootMountpoint be turned into a generator's UserConfig property, but without adding any flags. Thus the field becomes "for internal use" only.

And in such case you don't need the quirk part neither.

anatol (changes requested):

Issue #379, opened 2026-06-08
add defense against filesystem supplatation attacks


thePrivacyFanatic:
Note: this is technically solved right now in high security setups but is clunky and requires special effort.

Most is the initramfs generators available as of now have a vulnerability to a [filesystem supplantatin attack](https://oddlama.org/blog/bypassing-disk-encryption-with-tpm2-unlock/), in this attack a malicious LUKS volume is put on the disk as root which can (since only the UKI is verified at the time) unseal the TPM encapsulated key.

Possible solutions that I know of:
- systemd-cryptsetup measures all used volume keys into PCR#15 but even then the value has to be compared against a known held value in the initramfs.

- mkinitcpio allows adding runtime hooks and a hook that verifies a keyfile on the root and stops boot if mismatched can work (assuming the verification is hash based so the file is not up for grabs).

- booster allows use of a detached header which can implicitly sign it. (The only implemented solution but clunky).

The reason I think it is clunky is that decrypting from a detached header allows assuming the whole drive is encrypted, as such including header backups shouldn't work since the original header would also get "decrypted".
As such use of a detached header would also require backing the header up somewhere else and hinder legitimate recovery attempts.

An easier solution I propose is adding a header override option (in addition to red.luks.header or extending it) that would skip the used header without assuming it is part of the LUKS volume, and maybe a config option to extract and include the LUKS header of a volume in the initramfs. This would allow for less manual work (e.g. on header updates), but require some more configuration, so I may be overdoing it.

Regards.

anatol:
cc @pilotstew

pilotstew:
Did a little bit of research on this. TBH it's a tough issue but we can probably adopt some of the best practices seen elsewhere. I'm not sure how much time I'll have in the near future to tackle this but here are some ideas I've come across for hardening the auto-unlock path. The right fix of course is adding a second factor tpm+pin etc. But that's out of scope here. Hardening auto-unlock is doable but probably not bulletproof. I think this exposes greater risk for clevis than tpm. But in looking into this I uncovered a few other possible hardening targets I'll add to my todo's. 
  
  - PCR15 binding (single root). Seal the key to an unused/empty PCR15 the initramfs unseals while it's empty, then extends it before pivot, so the key can't be unsealed again once we've left the initramfs (systemd tpm2-measure-pcr / IncusOS's uninitialized-PCR15). Simple, but the first unlock dirties PCR15, so it really only fits a single volume.
[...]

PR #382, opened 2026-06-16
init: PCR 15 re-unseal latch for systemd-tpm2 unlock (#379)

pilotstew:
## Summary
Implements the runtime half of the TPM2 filesystem-supplantation hardening from #379. After a `systemd-tpm2` token is unsealed, booster extends PCR 15 with the systemd-compatible "system-identity" measurement before pivot, so the volume key cannot be re-unsealed for the rest of the boot — closing the TPM re-unseal oracle. Adds a `tpm2-measure-pcr=` control (crypttab and `rd.luks.options=`) and a manpage section documenting the secure-enrollment recipe.

## Background
A TPM attests the boot *path*, not the data it releases: with a passphrase-less, PCR-only enrollment a physical attacker can present a substituted root (or boot their own image) and have the TPM hand back the real key (oddlama, "Bypassing disk encryption with TPM2 unlock"; IncusOS CVE-2026-32606, fixed by binding the key to an uninitialized PCR 15).

## What it does
- `measureVolumeKeyToPCR15` extends PCR 15 with `HMAC(volume_key, "cryptsetup:" + name + ":" + uuid)` in every active PCR bank, byte-compatible with systemd-cryptsetup's `tpm2-measure-pcr=yes`, before `SetupMapper`/pivot.
- Engages automatically when the token's policy binds PCR 15 (the enrollment is the signal); `tpm2-measure-pcr=yes|no` overrides per device; a measurement failure aborts the unlock (fail closed).
- Extends **all** active banks — a policy satisfiable through an un-extended bank would otherwise be bypassable.
- The volume key is consumed via `luks.Volume.HMAC` (anatol/luks.go#17, merged) so the key never leaves the LUKS library — only the non-secret digest crosses into booster.

## What the testing settled — PCR 15 alone vs. co-binding
The thread's open question was whether PCR 15 contamination alone suffices (@thePrivacyFanatic) or whether PCR 11 co-binding is required. The characterization tests answer it empirically:
- PCR 15 alone closes the *within-boot* re-unseal oracle but is **not** a supplantation defense by itself — PCR 15 resets to zero on power-cycle, so an attacker who reboots into a malicious initrd re-unseals. `TestSupplantationPCR15WithoutPCR11Leaks` seals one secret to a PCR15-only policy and another to PCR 11+15, changes PCR 11 (a supplanted initrd), and shows the PCR15-only seal still unseals (leak) while PCR 11+15 refuses.
- `TestSupplantationCrossBootDenied` runs the real attack at the TPM level with an actual power-cycle: sealed to 7+11+15, an attacker boot that reproduces PCR 7 and a fresh PCR 15 but a different PCR 11 is denied, while the control boot (genuine PCR 11) succeeds — so **PCR 11 (the UKI measurement) is the load-bearing barrier and PCR 15 is the single-use latch.**

So the latch ships as @thePrivacyFanatic proposed (and composes with keyfile-cascaded secondary volumes as they noted), but the docs are explicit that PCR 11 co-binding is what actually resists supplantation.

## Tests
- e2e: `TestSupplantationPCR15Latch` — real boot, unseal at PCR 15 == 0, asserts the latch extends before pivot.
- premise/characterization (swtpm + tpm2-tools, executable threat-model): `…WithoutPCR11Leaks` (co-binding load-bearing), `…CrossBootDenied` (positive cross-boot), `…PCRBankIsolation` (all-banks mandatory), `…PCR15NonResettable`.
- unit: exact systemd HMAC formula + `xescape`, all-active-banks extend, fail-closed on TPM/bank errors, and the gating decision.

## On the PCR 11 enrollment pain (@thePrivacyFanatic, @invokekitty)
Literal PCR 11 binding already unseals on booster today (the token's PCR set is honored). The painful part raised in the thread — refreshing the PCR 11 value on every kernel update — is the literal-policy fragility documented in the new manpage section. Signed (authorized) PCR policy, which lets ukify re-sign the new PCR 11 each build instead of re-enrolling by hand, is the immediate follow-up PR.

## Scope
Runtime latch + control + docs only. The other #379 targets — TPM bus session encryption, post-unlock key cleanup/keyring/zeroing, header-embed, clevis, and signed PCR policy — are separate follow-ups (clevis has no upstream measurement mechanism; signed policy is next).

Addresses #379

### DIFF
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -151,7 +151,7 @@ Some parts of booster boot functionality can be modified with kernel boot parame
     * **Initramfs file** — an absolute path (e.g. `/etc/luks/root.hdr`) to a header file bundled into the initramfs at build time via `extra_files`.
     * **Raw block device** — a device path (e.g. `/dev/sdb`) where the LUKS header begins at byte offset 0. Booster waits for the device to appear and passes it directly to cryptsetup without mounting.
     * **File on a separate device** — `$path:$deviceref` where `$deviceref` is `UUID=...`, `LABEL=...`, `PARTUUID=...`, or `PARTLABEL=...`. Booster mounts the device read-only, reads the header file, then unmounts before unlocking.
- * `rd.luks.options=opt1,opt2` a comma-separated list of LUKS flags. Supported options are `discard`, `same-cpu-crypt`, `submit-from-crypt-cpus`, `no-read-workqueue`, `no-write-workqueue`. `token-timeout=<duration>` sets how long to wait for hardware tokens (FIDO2, TPM2) before also prompting for a keyboard passphrase. Accepts a decimal number followed by a unit (`s`, `m`, `h`), or a bare integer treated as seconds. Default is 30 s.
+ * `rd.luks.options=opt1,opt2` a comma-separated list of LUKS flags. Supported options are `discard`, `same-cpu-crypt`, `submit-from-crypt-cpus`, `no-read-workqueue`, `no-write-workqueue`. `token-timeout=<duration>` sets how long to wait for hardware tokens (FIDO2, TPM2) before also prompting for a keyboard passphrase. Accepts a decimal number followed by a unit (`s`, `m`, `h`), or a bare integer treated as seconds. Default is 30 s. `tpm2-measure-pcr=yes` or `tpm2-measure-pcr=no` forces or suppresses the PCR 15 re-unseal latch for the device (see **TPM2 unlock and the PCR 15 latch**); the command-line value takes precedence over crypttab.
     Note that booster also supports LUKS v2 persistent flags stored with the partition metadata. Any command-line options are added on top of the persistent flags.
  * `rd.modules_force_load` a comma-separated list of extra kernel modules which should be force loaded.
  * `resume=$deviceref` device reference to suspend-to-disk device.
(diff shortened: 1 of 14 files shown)


anatol:
The code looks good to me.

@thePrivacyFanatic @invokekitty is there any chance you can try it to see if the functionality matches your expectations?

thePrivacyFanatic:
i have tried, but i can't seem to install booster-git from AUR (even without editing the PKGBUILD to pull pr/tpm-pcr15-latch) and  `go build./...` failed with an error as well...
would like some guidance on how to build it.

pilotstew:
Build the two binaries explicitly:
```
git clone https://github.com/pilotstew/booster && cd booster
git checkout pr/tpm-pcr15-latch
go build -o booster ./generator
go build -o booster-init ./init
```

You may need the fido plugin if your `/etc/booster.yaml` sets `enable_fido2: true` or a crypttab entry uses `fido2-device=` — in that case the `booster build` below would abort with `fido2 plugin …: no such file` until you build it and place it next to the init binary. If so, in addition to the above:
```
go build -buildmode=plugin -o fido2plugin.so ./init/fido2plugin/
```

The easiest way to watch the latch fire without touching your real system is the integration test — it boots a real image in QEMU against a software TPM and needs none of the above (needs `swtpm`, `qemu`, and sudo for the test disk's loopback/cryptsetup setup):
```
cd tests && go test -run TestSupplantationPCR15Latch -v
```
You'll see `recovered password from systemd-tpm2 token` → `PCR15 re-unseal latch engaged` → `Hello, booster!`.

To try it against a real enrollment, build an image with the built init from above and enroll a TPM2 token bound to PCR 7+11+15:
```
sudo ./booster build --init-binary ./booster-init /boot/booster-test.img
sudo systemd-cryptenroll --tpm2-device=auto --tpm2-pcrs=7+11+15 /dev/sda2
```
As mentioned in the PR, PCR 15 alone wasn't enough in my testing — PCR 11 is what fixes it: it measures the kernel+initrd (via the UKI), so a swapped or malicious boot image yields a different PCR 11 and the TPM refuses to release the key.

Boot via a new boot entry in grub, systemd-boot, limine, etc. — so you can still log in with your current img/boot should something fail with the new entry. Include `booster.log=debug` and look for `PCR15 re-unseal latch engaged`. 

If the above doesn't work, paste the error here.

I'm also looking for feedback on the new "TPM2 unlock and the PCR 15 latch" section in the manpage, if it's unclear let me know.

pilotstew:
The signed PCR11 follow up is  up for preview at https://github.com/pilotstew/booster/tree/pr/tpm-signed-pcr-policy. I rewrote the manpage sections on this I think it reads much more clearly and I included a _hopefully_ easy read example for the full config.

It builds on this PR but I can open if you'd prefer.

invokekitty:
Have tested it on my machine (through the tpm-signed-pcr-policy branch) and it works as expected

anatol:
Thank you very much for the great work again. The changes have landed repo.

pilotstew:
> Have tested it on my machine (through the tpm-signed-pcr-policy branch) and it works as expected

Excellent. With this PR merged I can push that branch now. Thank you for testing it out ahead of time.

PR #383, opened 2026-06-20
init: signed (authorized) PCR policy for systemd-tpm2 unlock

pilotstew:
Builds on #382 (merged) and is the piece that addresses #379 in principle.

#382 added the PCR 15 re-unseal latch, which makes each TPM2 unseal single-use within a boot. But the latch alone is not the supplantation defense: PCR 15 resets to zero every power cycle, so an attacker who reboots into a substituted kernel or initrd sees a fresh PCR 15 and can unseal again. The barrier that actually stops that is PCR 11 — the UKI measurement of kernel + initrd + command line. A substituted kernel or initrd produces a different PCR 11 and the TPM withholds the key. This PR adds practical PCR 11 binding, so it is the change that closes the supplantation gap #379 is about; the remaining items from that issue (TPM bus-session encryption, in-memory key cleanup) are separate hardening.

The reason PCR 11 was not already usable is that it cannot be bound *literally*: systemd advances PCR 11 through boot phases, so the value at unlock differs from any value enrollment could record, and every kernel or initrd update changes it — a literal binding forces re-enrollment on every update, and (as raised in #379) makes enrollment supply the UKI hash by hand. systemd solves this with a signed (authorized) PCR policy: `ukify`/`systemd-measure` re-sign the new PCR 11 value on each build, and the token authorizes any PCR 11 state the trusted key has signed. Booster did not evaluate signed policies, so PCR 11 binding was effectively unavailable — including the PCR 11 + PIN configuration one user noted in #379 was keeping them on systemd/mkinitcpio.

This PR adds signed-policy support to the `systemd-tpm2` unlock path:

- `TPM2_PolicyAuthorize` over a `systemd-cryptenroll --tpm2-public-key` token, verifying the `.pcrsig` signature `ukify`/`systemd-measure` produce, so PCR 11 can be bound and the same sealed blob keeps unsealing across kernel and initrd updates with no re-enrollment.
- Composition in systemd's order — `TPM2_PolicyAuthorize` (signed PCR 11), then `TPM2_PolicyPCR` (literal PCR 7 and 15), then `TPM2_PolicyAuthValue` (PIN) — so one token can carry signed PCR 11, live PCR 7, the PCR 15 latch, and a PIN together.
- The `enter-initrd` boot-phase measurement extended into PCR 11 just before the unseal (booster runs no `systemd-pcrphase`), so the live value matches the signed one.
- A `tpm2-signature=<path>` crypttab / `rd.luks.options=` option. When unset, booster reads the signature systemd-stub unpacks from the UKI at `/.extra/tpm2-pcr-signature.json`; a non-UKI boot supplies it explicitly.
- Fail-safe rejection of sharded and pcrlock-bound tokens (booster can satisfy neither), with a clear message instead of a cryptic unmarshal error.

Verification:

- swtpm unit tests covering the signed unseal end to end: PolicyAuthorize over a real RSA-signed policy, survival across a PCR 11 change, literal PCR 7/15 composition, the all-zero PCR 15 latch binding, PIN composition, and the `enter-initrd` barrier.
- An integration test (`TestSystemdTPM2SignedPolicy`) that boots an image whose token is a real `systemd-cryptenroll --tpm2-public-key` enrollment, with the signature computed from real systemd artifacts.
- An end-to-end hardware boot: Secure Boot + UKI, unlocking from a token bound to PCR 7 + signed PCR 11 + the all-zero PCR 15 latch + a PIN.
- Independently tested on real hardware by @invokekitty against this branch, in the PCR 11 + PIN configuration they rely on, confirmed working as expected (#382).

Docs: the manpage's TPM2 section is rewritten to walk the bindings in boot order (Secure Boot → PCR 7, kernel+initrd → signed PCR 11, the unseal, the PCR 15 latch, the leave-initrd forward-lock), with a worked headless-with-SSH-recovery enrollment example.


anatol:
Fantastic! Thank you very much @pilotstew! And thank you for your feedback @invokekitty!

PR #385, opened 2026-06-25
init: make the PCR 15 re-unseal latch default-on (opt-out)

pilotstew:
## Summary

Follow-up to the recently-merged PCR 15 re-unseal latch. It was opt-in — it fired only when the opened volume's `systemd-tpm2` token bound PCR 15, or when forced with `tpm2-measure-pcr=yes`. But that token lives on the data device the attacker controls, so a tokenless fake volume (the oddlama filesystem-confusion case) just skips it, leaving PCR 15 at zero and the volume key re-unsealable after pivot. The latch should fire whenever a LUKS root is unlocked, not gate on the on-disk token.

This makes it default-on. `volumeKeyLatchMode` now extends PCR 15 whenever a TPM is present, even with no token — best-effort, so a glitchy or absent TPM never blocks a passphrase boot. The token-bound and `tpm2-measure-pcr=yes` cases stay fail-closed as before; `tpm2-measure-pcr=no` opts out.

The PCR 11 boot-phase words had the same token-gating: booster extended `enter-initrd`/`leave-initrd` only when a signed-PCR 11 token was processed. They are now extended unconditionally at `switch_root`, matching `systemd-pcrphase-initrd` (which measures them regardless of unlock), so the PCR 11 phase value is consistent across boots whether or not a token was used.

The latch only helps a key that actually binds PCR 15 (the recommended 7+11+15) — noted in the manpage. The `leave-initrd` PCR 11 forward-lock does not cover the filesystem-confusion case on its own, since `systemd-measure`'s default phases also sign the `leave-initrd` state; `init/tpm_signed_test.go` shows the difference against the real unseal path.

## Testing

Unit (`TestVolumeKeyLatchMode` plus the existing latch tests), swtpm signed-policy tests against a default-phase signature (`TestPCR15LatchDeniesAtLeaveInitrd`, `TestLeaveInitrdForwardLockDefeatedByDefaultPhaseSignature`), and the QEMU boots (`TestSystemdTPM2`, `TestSystemdTPM2SignedPolicy`, `TestSupplantationPCR15Latch`, `TestSystemdRecovery`) — the defensive latch engages without blocking handoff. Also folds in a couple of now-redundant tests from the earlier work.


anatol:
Thank you for this change. It looks fantastic. But as usual it would be get more eyes on it.

@invokekitty @thePrivacyFanatic is there any chance you can help to cross-verify this feature?

anatol (approved):

thePrivacyFanatic:
> Thank you for this change. It looks fantastic. But as usual it would be get more eyes on it.
> 
> @invokekitty @thePrivacyFanatic is there any chance you can help to cross-verify this feature?

i guess i'm late, but as a feature this is needed to make the most out of the PCR#15 unseal.

Edit: it does also build fine and unseal automatically as it should for me

anatol:
Thank you @pilotstew for another great contribution.

PR #387, opened 2026-07-03
init/console: configurable password echo cycle (asterisks/silent/plaintext)

adelmonte:
### What this PR does

Makes the password prompt's keystroke feedback configurable, and implements the plaintext reveal the manpage already describes.

Since the raw-mode reader landed (#355), the manpage has said:

> **Tab** — toggle visibility of the typed characters (asterisks ↔ literal).

but the literal mode was never wired up — the `masked` bool actually toggles asterisks ↔ *hidden*. This PR turns that two-state bool into a configurable mode cycle and closes the doc/code gap:

- **Three echo modes**: `asterisks` (one `*` per codepoint; the default), `silent` (no feedback at all, sudo-style), `plaintext` (literal characters, for hunting down a typo).
- **`password_echo`** is an ordered, comma-separated list of modes (same syntax as `modules`/`extra_files`): the prompt starts in the first listed mode and **Tab cycles through the list in order, wrapping**. The default `asterisks,silent,plaintext` keeps the historical first-press behavior (asterisks → hidden) and makes plaintext reachable only deliberately — never on the way to hiding.
- **A single entry** (`password_echo: silent`) pins the prompt to one mode — Tab becomes a no-op (some users prefer the prompt to have exactly one behavior).
- **Omitting `plaintext`** (`password_echo: silent,asterisks`) allows toggling between the hidden modes while making the typed password unrevealable at the prompt.
- Applies to every prompt that goes through the shared reader: LUKS passphrase, TPM2 PIN, FIDO2 PIN.

**Defaults are unchanged**: with no config, the prompt echoes asterisks and Tab's first press hides them, exactly as today.

### Motivation

The asterisk echo added in #355 (requested in #305) is a regression for users who preferred the old sudo-style silent prompt — there is currently no way to get it back except pressing Tab on every boot. Listing `silent` first restores it persistently, while keeping the #305 behavior as the default. Expressing the cycle as an ordered list (rather than a fixed cycle plus a toggle flag) also covers what a fixed order can't: whatever mode you start in, you can order the list so plaintext is never one accidental Tab away — or leave it out entirely.

### Implementation notes

- `consolePrompt` now tracks the exact painted feedback (`echoed string`) instead of an asterisk count, so the log-over-prompt redraw paths (`consolePrintWithPromptRedraw`, `statusMessage`) repaint silent and plaintext prompts correctly instead of assuming asterisks.
- The generator validates `password_echo` at build time (known modes only, no duplicates, no empty elements) and fails loudly on a bad value; init re-parses the same syntax defensively and falls back to the default cycle with a warning — a cosmetic setting must never fail a boot.
- One config key covers startup mode, cycle order, membership, and pinning — there is no second flag to keep consistent with it.
- Erase paths assume one terminal cell per codepoint; a double-width glyph echoed in plaintext leaves half a cell behind on backspace (cosmetic only, buffer stays correct — same assumption the asterisk path has always made, noted in a comment).

### Tests

New PTY-harness tests capture `os.Stdout` (where the reader echoes; the pty slave is input-only) and assert on what a user would actually see: silent emits nothing, plaintext echoes literally and erases on backspace, Tab repaints the buffer across mode transitions, a hidden-only cycle (`silent,asterisks`) never paints the typed characters, a single-mode cycle swallows Tab without typing it into the password, and the redraw path repaints the painted feedback verbatim. Plus table tests for the list parser (ordering, whitespace trimming, rejection of unknown/duplicate/empty modes) and generator-side validation.

`go test ./init/ ./generator/` — all reader/scanner/config tests pass (the environment-dependent blkinfo/image tests fail identically on pristine master).

### Docs

`docs/manpage.md` — `password_echo` documented, the example config updated, and the Password entry section now describes the real behavior; `CHANGES.md` entry added.

### DIFF
--- a/init/console_input.go
+++ b/init/console_input.go
@@ -89,6 +89,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"slices"
 	"strings"
 	"sync"
 	"unicode/utf8"
@@ -106,7 +107,7 @@ const (
 	keyEnter                     // CR / LF — end of input
 	keyEOF                       // Ctrl+D (0x04) — end-of-input on empty buffer; no-op otherwise
 	keyBackspace                 // BS (0x08) or DEL (0x7f) — erase last char
-	keyTab                       // Tab (0x09) — toggle masking
+	keyTab                       // Tab (0x09) — cycle the password echo mode
 	keyKillLine                  // Ctrl+U (0x15) — erase entire line
 	keyKillWord                  // Ctrl+W (0x17) — erase back to last whitespace
 )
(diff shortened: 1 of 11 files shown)


anatol:
Cc @pilotstew

adelmonte:
Heads-up for reviewers: I reworked this before review and force-pushed. The earlier revision had a fixed Tab cycle plus a separate `password_echo_toggle` on/off flag; that design couldn't express "start silent without plaintext being one Tab away" or "toggle between the hidden modes only". `password_echo` now takes an ordered, comma-separated list of modes instead — first entry is the startup mode, Tab cycles the list, a single entry pins the prompt, and omitting `plaintext` makes the password unrevealable. One key instead of two, same list syntax as `modules`/`extra_files`, and the default remains behavior-identical to master. The PR description has been rewritten to match.

init/plymouth.go
```diff
@@ -218,7 +218,7 @@ func statusMessage(msg string) {
 	} else if msg != "" {
 		consoleMu.Lock()
 		if consolePrompt.active && !promptVolumeUnlocked() {
-			consolePrint("\n\n" + msg + "\n\n" + consolePrompt.text + strings.Repeat("*", consolePrompt.asterisks))
+			consolePrint("\n\n" + msg + "\n\n" + consolePrompt.text + consolePrompt.echoed)
```
pilotstew: Same scrollback leak:

```
consolePrint("\r\x1b[K" + "\n\n" + msg + "\n\n" + consolePrompt.text + consolePrompt.echoed)
```

init/console_input.go
```diff
@@ -518,7 +588,7 @@ func consolePrintWithPromptRedraw(msg string) {
 	consoleMu.Lock()
 	defer consoleMu.Unlock()
 	if consolePrompt.active && !promptVolumeUnlocked() {
-		consolePrint(msg + "\n" + consolePrompt.text + strings.Repeat("*", consolePrompt.asterisks))
+		consolePrint(msg + "\n" + consolePrompt.text + consolePrompt.echoed)
```
pilotstew: This scrolls the live prompt line (which shows the literal passphrase in `plaintext` mode) up into scrollback intact . The leading `\n` moves past it without erasing it. Erase the line first:

```
consolePrint("\r\x1b[K" + msg + "\n" + consolePrompt.text + consolePrompt.echoed)
```

That wipes the live line in place before scrolling, so only the status text reaches history; the reprint below stays verbatim so the plain text view is unchanged.

pilotstew (commented):
Two minor security concerns on reprompt.

init/console_input.go
adelmonte: Fixed in e911a52. One deviation from the literal snippet: the erase is emitted via `consoleEcho` rather than glued into the `consolePrint` string, so `-tags test` builds (where `consolePrint` targets `/dev/kmsg`) don't write raw escape bytes into the kernel log. While touching this I also noticed the repaint itself would log the echoed feedback — the literal passphrase in plaintext mode — to kmsg in those test builds, so the repaint now skips the feedback under `quirk.TestEnabled`. Both redraw paths are covered by tests asserting the erase lands before the status text.

init/plymouth.go
adelmonte: Fixed in e911a52, same treatment as the console_input.go path (erase via `consoleEcho`, feedback repaint guarded in `-tags test` builds). Amusingly `statusMessage`'s doc comment already claimed "the current line is erased" — the code just never did it; now it does.

pilotstew: Correct, I had intended it and didn't realize until I saw your code snippet that it wasn't live. The risk was minimal before but more important to fix now.

pilotstew:
Thanks @adelmonte

anatol:
Thank you @adelmonte!

Could you please amend the `init/console:` commits together thus changes like formatting are not shown in the final history?

@pilotstew are changes look good and ready for the merge into mainline?

adelmonte:
Done — folded the fix commit into the original commits as requested: the erase-before-scroll change and its `init/config.go` formatting fallout into the `init/console:` commit, and the stray `generator`/test formatting hunks into their own home commits rather than riding along in the squash. Still four commits, and the final tree is byte-identical to what was reviewed.

pilotstew:
Looks good to me

PR #390, opened 2026-07-05
docs/manpage: rd.luks.header takes the LUKS superblock UUID

pilotstew:
The rd.luks.header= manpage entry currently documents the left-hand value as a partition UUID. It is the LUKS superblock UUID. The same value rd.luks.uuid takes and what cryptsetup luksUUID --header prints. Booster parses it into a filesystem-UUID device reference and matches it against the UUID read out of the detached header; a PARTUUID there builds a reference that can never equal the header's UUID, so the device never unlocks. This restores the correct identifier while keeping the accurate observation that a detached-header data device exposes no UUID of its own.

Two related clarifications:

The cmdline/crypttab merge reconciles the two sources only by LUKS UUID, so a crypttab entry that names the same device by a partition identifier is treated as a distinct mapping and its options do not merge. This matches systemd-cryptsetup.

Multiple detached headers need each data device named explicitly in crypttab (PARTUUID/PARTLABEL/LABEL/path) with its own header=. rd.luks.header= supplies only the header, so with several headerless data devices present booster cannot tell which device pairs with which header and the pairing falls back to device arrival order.

Booster currently doesn't handle the systemd style for command-line detached-header. Systemd names the data device on the cmdline with rd.luks.data=$UUID=$DEV (the analogue of crypttab's encrypted-device field), whereas booster has no rd.luks.data= equivalent and accepts only the header on the command line. So crypttab is currently the only way to pin the data device for a multi-header setup. Adding rd.luks.data= would be a good addition.

### DIFF
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -147,7 +147,7 @@ Some parts of booster boot functionality can be modified with kernel boot parame
  * `rd.luks.uuid=$UUID` UUID of the LUKS partition where the root partition is enclosed. booster will try to unlock this LUKS device.
  * `rd.luks.name=$UUID=$NAME` similar to rd.luks.uuid parameter but also specifies the name used for the LUKS device opening.
  * `rd.luks.key=$UUID=$PATH` absolute path to a keyfile in the initrd/initramfs which can be used to unlock the device identified by UUID, if this file does not exist or fails to unlock it will fall back to a password request.
- * `rd.luks.header=$PARTUUID=$PATH` detached LUKS header for the device identified by `$PARTUUID` (notice that no filesystem UUID is shown when the header is detached as it is stored in the header. `$PATH` can take three forms:
+ * `rd.luks.header=$UUID=$PATH` detached LUKS header for the device identified by `$UUID` — the LUKS superblock UUID, the same value `rd.luks.uuid` takes and what `cryptsetup luksUUID --header $PATH` prints. Because that UUID lives *in the header*, a device with a detached header exposes no LUKS or filesystem UUID of its own; identify it by this header UUID, not by the data partition's `PARTUUID`. `$PATH` can take three forms:
     * **Initramfs file** — an absolute path (e.g. `/etc/luks/root.hdr`) to a header file bundled into the initramfs at build time via `extra_files`.
     * **Raw block device** — a device path (e.g. `/dev/sdb`) where the LUKS header begins at byte offset 0. Booster waits for the device to appear and passes it directly to cryptsetup without mounting.
     * **File on a separate device** — `$path:$deviceref` where `$deviceref` is `UUID=...`, `LABEL=...`, `PARTUUID=...`, or `PARTLABEL=...`. Booster mounts the device read-only, reads the header file, then unmounts before unlocking.
@@ -247,7 +247,14 @@ same device, the cmdline takes precedence for the device reference and
 mapper name, and for any security option (keyfile, header, tries,
 token-timeout, …) it sets explicitly; the crypttab entry supplies only
 the options the cmdline left unset. Crypttab entries for devices not
-covered by `rd.luks.*` are appended as new mappings.
+covered by `rd.luks.*` are appended as new mappings. Two entries "cover
+the same device" only when they name it the same way: a crypttab entry
+merges with a `rd.luks.*` parameter only if its device field is the same
+LUKS UUID (`UUID=<luksUUID>`). A crypttab entry that instead uses a
+partition identifier (`PARTUUID=`, `PARTLABEL=`, `LABEL=`, or a path) is
+treated as a distinct device and appended as its own mapping, so its
+options do **not** merge onto the cmdline entry — `systemd-cryptsetup`
+reconciles the two sources by LUKS UUID the same way.
 
 Booster-specific behaviour for selected options:
 


anatol:
cc @thePrivacyFanatic for a review

pilotstew:
Here's the whole picture tied together — the three IDs, which config field wants which, how each is physically read, and the two different ways the header and device pair up at boot.

### The three IDs (one line each)

- **LUKS UUID** = the lock's ID. Lives in the LUKS header.
- **PARTUUID** = the slot's ID. Lives in the GPT partition table.
- **Filesystem UUID** = the contents' ID. Lives in the fs superblock, only visible after unlock.

### The config confusion — which field eats which ID

| Config field | Wants | Common wrong guess |
|---|---|---|
| `rd.luks.uuid=` / `rd.luks.name=` | LUKS UUID | — |
| `rd.luks.header=$UUID=$path` (left side) | LUKS UUID | ❌ PARTUUID (the recent regression) |
| crypttab device field | LUKS UUID (`UUID=…`) or a partition ref (`PARTUUID=`/`PARTLABEL=`/`LABEL=`/path) | — |
| `root=` | Filesystem UUID (or `/dev/mapper/name`) | ❌ the LUKS UUID |

Why it's confusing: on an attached-header device, `blkid /dev/nvme0n1p2` prints the LUKS UUID and the PARTUUID on the same line (`UUID="fc51…" TYPE="crypto_LUKS" PARTUUID="0cf4…"`). Detach the header and that same `blkid` prints nothing but the PARTUUID — the LUKS UUID walked off into the header file. So the "obvious" UUID you'd copy changes depending on attached vs detached, which is exactly the trap.

### How each ID is physically accessed (read)

| ID | Read from | Terminal | Booster code |
|---|---|---|---|
| LUKS UUID | the header — attached: partition byte `0xa8`; detached: the header file | `cryptsetup luksUUID <hdr>` | `probeLuksHeader` reads magic `LUKS\xba\xbe` then the UUID at `0xa8` (blkinfo.go:274, :284-289) |
| PARTUUID | the GPT table on the disk (not the partition body) | `blkid -s PARTUUID -o value /dev/…` | `resolveGptRef` reads the parent GPT's partition entries (deviceref.go:76) |
| Filesystem UUID | the fs superblock inside the unlocked mapper | `blkid /dev/mapper/cryptroot` | normal `readBlkInfo` after unlock |

Note the LUKS UUID is never read off the headerless data device — there's nothing there to read. That single fact drives everything below.

### How the header and device find each other at boot

Both paths start the same: config is parsed into a `luksMapping{ ref, header }`, then udev fires `addBlockDevice → readBlkInfo`. The detached-header data device has no LUKS magic, so it comes back `format==""` with `blk.uuid == nil` (main.go:300). From there the two config styles pair differently:

**Path A — cmdline `rd.luks.header=$LUKSUUID=$path`** (ref is a LUKS-UUID ref, `refFsUUID` — luks.go:1884)

```
device arrives (format=="", uuid=nil)
  → blk.matchesRef(m.ref)  → FALSE   (nil uuid can't equal the LUKS UUID)     main.go:335
  → read the header:  hdrBlk = probeLuksHeader(m.header)                       main.go:341
  → hdrBlk.matchesRef(m.ref)  → does the HEADER's UUID == the declared UUID?   main.go:345
      ↑ this tests the header against itself — it never looks at the device
  → yes → adopt THIS device (whatever it is)                                   main.go:346
```

So the header validates itself against the declared UUID, and the device is just "whichever headerless one showed up." That's why the pairing is device-blind and can't disambiguate multiple headerless devices — the match doesn't depend on the device at all.

**Path B — crypttab `name PARTUUID=X none header=$path`** (ref is a partition ref, from `parseDeviceRef` — crypttab.go:52)

```
GPT disk processed  → resolveGptRef rewrites m.ref: PARTUUID=X → /dev/nvme0n1pN   main.go:458
device arrives (the partition with PARTUUID=X)
  → blk.matchesRef(m.ref)  → TRUE   (matches the specific device by path)         main.go:335
  → adopt THAT specific device; header comes along via acquireHeader              main.go:338
```

Here the device is pinned by PARTUUID and the header is simply attached to it — unambiguous, works for any number of headers.

Then, identically for both, the header is applied to the chosen device:

```
luksOpen(dev, m)  →  headerPath = acquireHeader(m)  →  luks.OpenWithHeader(dev, headerPath)   luks.go:1519, 1526, 1532
```

cryptsetup applies the header's key material to `dev` and does not verify the two belong together — if the wrong device was adopted, it just decrypts to garbage.

### The one-sentence takeaway

- **Cmdline** finds the pair through the header (header self-validates by LUKS UUID; device is incidental) → device-blind, single-header-only.
- **crypttab** finds the pair through the device (PARTUUID pins the exact partition; header rides along) → unambiguous, multi-header-capable.

That asymmetry is the entire reason the guidance is "use crypttab with `PARTUUID=` for detached headers," and why the missing `rd.luks.data=` (systemd's way to pin the device on the cmdline) is the real gap.

thePrivacyFanatic:
Pilotstew is far better than me at go so if he said that's how the code works I'd trust him, though I myself can't really review that claim.
As such, the PR seems fine AFAIK.
But it is strange as program behavior, which is why I wrongly believed it would work the other way.
If it is would you like me to edit and reopen my issue about it?

pilotstew:
On my todo list for a bit now is rd.luks.options=$UUID=... This would allow better per device kernel cmdline handling and has been the source of a few issues. The quick fix for now is adding rd.luks.data=. It's small, I'm about finished with testing and it gets you what your asking for on the cmdline. My focus has been on crypttab, I think it's better suited for these types of operations.

thePrivacyFanatic:
Thanks, no need to hurry though as it does seem crypttab is already a really good option for advanced setups (as long as you disregard the big note that tells you root doesn't belong there) so this is just a UX improvement for those "middle complexity" setups

anatol:
Thank you everyone!

pilotstew:
> Thanks, no need to hurry though as it does seem crypttab is already a really good option for advanced setups (as long as you disregard the big note that tells you root doesn't belong there) so this is just a UX improvement for those "middle complexity" setups

This is I think a common misconception. Root has been flagged here simply because /etc/crypttab is loaded after the initramfs, therefore root would not be found. Hence the warning. However, crypttab's x-initrd.attach is still fairly new and specifically includes that line in the initramfs. Several distro's and systemd now explicitly recommend including root in crypttab with x-initrd.attach. This is how I have my root defined. Are you referring to the arch wiki's flag here?

thePrivacyFanatic:
No you got it right I've just meant that the x-initrd.attach option is a really simple and swift way and I guess they warn against it since it's new and not all initramfs generators use it (or at least suggest using it) yet

pilotstew:
You're correct, there is a split between a few different methods currently. x-initrd.attach is the systemd convention, mkinitcpio uses a seperate /etc/crypttab.initramfs and I believe debian pulls automatically from fstab. Fragmented conventions makes it more difficult. In fact my inital build of crypttab support used the crypttab.initramfs spec but overall I thought this was the better option

anatol:
> there is a split between a few different methods currently. x-initrd.attach is the systemd convention, mkinitcpio uses a seperate /etc/crypttab.initramfs and I believe debian pulls automatically from fstab. Fragmented conventions makes it more difficult. 

It is pretty useful information. Would it be possible to include it to the documentation, probably to manpage, section 'Gotchas' or something like that.

pilotstew:
Sure, I'll get this rd.luks.data PR out first then look at the manpage.

PR #391, opened 2026-07-06
init: rd.luks.data= to pin the data device for detached LUKS headers

pilotstew:
Booster pairs a detached header with its data device two different ways depending on where the volume is declared, and they are not symmetric:

- **Kernel command line** (`rd.luks.header=$UUID=$path`): the data device has no identifier of its own, so booster reads the header and checks that the UUID inside equals the declared `$UUID`. That validates the header against the mapping — it never looks at the device. It's device-blind: fine for one detached-header volume, but with several it cannot tell which device belongs to which header, and the pairing falls back to device arrival order.
- **crypttab** (`name PARTUUID=… none header=$path`): the device field pins the exact partition and the header rides along. Unambiguous, and it scales to any number of detached-header volumes.

**systemd parity.** `rd.luks.data=$UUID=$deviceref` mirrors systemd-cryptsetup's parameter of the same name: keyed by the LUKS superblock UUID, naming the encrypted-data device, and accepting the same fstab-style specifiers (`UUID=`, `PARTUUID=`, `LABEL=`, `PARTLABEL=`, or a path). systemd documents it as "the analogue of the second crypttab field (encrypted-device)", specifically for detached-header data devices. The header is named differently: systemd puts it in `rd.luks.options=$UUID=header=$path`, whereas booster uses its own `rd.luks.header=$UUID=$path`.

cryptsetup does not verify that a header and a data device belong together — a wrong adoption just decrypts to garbage — so pinning the device is what makes multi-volume detached-header unlock reliable on the command line.

**Implementation.**
- Parse `rd.luks.data=$UUID=$deviceref` into a per-mapping data-device reference (`init/cmdline.go`).
- When a headerless device arrives and its mapping carries a data-device pin, match the device against that reference and skip the device-blind header fallback; `PARTUUID=`/`PARTLABEL=` refs resolve through the existing GPT path (`init/main.go`).
- The parse case also removes a silent-ignore: an unrecognized `rd.luks.data=` previously fell through to the module-parameter path and was discarded without warning.

**Compatibility.** With no `rd.luks.data=`, single detached-header behaviour is unchanged.

**Tests.**
- Unit test for the parser.
- Integration: a single detached-header volume pinned by device path.
- Integration: two detached-header volumes, the real root on the second-arriving device, each pinned by `rd.luks.data=`. Without the pins the same scenario mis-pairs and root never unlocks — a genuine regression signal for the multi-device case.


anatol:
cc @thePrivacyFanatic

thePrivacyFanatic:
I get from the manage commit that the way this works is I put both 
`rd.luks.header=uuid1=header1` and `rd.luks.data=uuid1=deviceref1` and it works like
`cryptsetup luksOpen device1 --header header1`
If I got this right I think it's a good solution.
I also guess it would work well with options like rd.luks.name if they all use the same system of LUKS UUID mappings, so that's very nice.
I did previously think that rd.luks.* is limited to exactly 1 volume so I'd like to make sure i was wrong about it before making a PR to clarify that for others.

pilotstew:
Yes. Again this is the quick functional but verbose fix. Rd.luks.options would allow for rd.luks.options.$UUID convention but for now booster only accepts global rd.luks.options=op1,op2

anatol:
Thank you @pilotstew !

PR #393, opened 2026-07-06
docs/manpage: fix the luksUUID example and document crypttab initramfs conventions

pilotstew:
Two manpage clarifications following the detached-header discussion.

The `rd.luks.header=` entry says the value is what `cryptsetup luksUUID --header $PATH` prints. That form errors — `--header` sets the detached-metadata location and still expects a data device as its positional argument (`cryptsetup luksUUID: requires <device> as arguments`). Reading the UUID out of a detached header is done by running cryptsetup on the header file directly: `cryptsetup luksUUID $PATH`.

Document how crypttab entries reach the initramfs, and clarify that the root LUKS device belongs in crypttab when marked `x-initrd.attach` — the common "root doesn't belong in crypttab" caution concerns the post-boot `/etc/crypttab`, which is read too late for root. Note the cross-generator split: booster follows systemd's `x-initrd.attach`, mkinitcpio uses a separate `/etc/crypttab.initramfs`, and Debian's cryptsetup-initramfs uses an `initramfs` option and auto-detects root from `/etc/fstab`. Booster recognizes only `x-initrd.attach`.


anatol:
thank you for the documentation improvements @pilotstew !

PR #402, opened 2026-07-14
init: password-gate the emergency shell

pilotstew:
booster's emergencyShell() execs `busybox sh -I` as root in the initramfs on the boot-failure path whenever busybox has been added via extra_files, with no authentication. An attacker who can force boost() to fail (withhold or present an unmountable root) reaches an unauthenticated root shell in the trusted, pre-pivot boot phase.

Add an opt-in emergency_shell_password config holding an argon2id PHC string. When set, booster prompts for the password before starting the shell and reboots after three wrong attempts; when unset, the shell starts unauthenticated as before, so existing images are unchanged.

verifyPassword parses the systemd/argon2-CLI PHC format (unpadded base64) and compares in constant time. authorizeEmergencyShell fails closed: a read error or a malformed configured hash denies the shell rather than granting it. The hash is minted externally with the argon2 CLI, so there is no new dependency (x/crypto/argon2 is already vendored) and no new subcommand. Documented in the manpage.

### DIFF
--- a/init/main.go
+++ b/init/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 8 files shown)


init/main.go
```diff
@@ -1351,15 +1352,21 @@ func enableLocalEcho() error {
...
-			warning("Failed to enable local echo: %v", err)
-		}
-
-		if err := unix.Exec("/usr/bin/busybox", []string{"sh", "-I"}, nil); err != nil {
-			severe("Unable to start an emergency shell: %v", err)
-		}
+	if _, err := os.Stat("/usr/bin/busybox"); os.IsNotExist(err) {
+		return
+	}
+	ok := authorizeEmergencyShell(config.EmergencyShellPassword, func() ([]byte, error) {
```
anatol: I would redo it to make a bit more readable flow as 

```go
authFn := func() ([]byte, error) {
	return readPassword(context.Background(), "Emergency shell password: ", "")
}
if ok := authorizeEmergencyShell(config.EmergencyShellPassword, authFn, 3); !ok {
	return
}
```

anatol (approved):

init/main.go
pilotstew: Done and thank you!

thePrivacyFanatic:
Many thanks!
Now I can have recovery without sabotaging the security.
Alternatively, previously we had to be cautious to not set busybox for this exact reason and that's nice as well!

PR #404, opened 2026-07-18
init: zero LUKS passphrases and TPM2 PINs from memory after use

pilotstew:
## Summary

Secret material (LUKS passphrases, TPM2 PINs, unsealed volume keys) currently lingers in the initramfs's memory after it has served its purpose. Go does not scrub freed slices, and unless the kernel is built with `CONFIG_INIT_ON_FREE_DEFAULT_ON=y` those pages keep the plaintext until they happen to be reused — a cold-boot / RAM-remanence exposure that outlives the unlock. 

No real functional changes...This adds a small `wipe()` primitive and scrubs in-process `[]byte` secrets at the point they stop being needed.

## What gets wiped

- The passphrase cache, at both boot exits — the `switchRoot` handoff (via `cleanup()`) and the emergency shell. The root passphrase no longer survives into the pivoted system or an interactive rescue shell.
- Failed keyboard passphrase attempts, scrubbed at the retry site (no one owns them).
- TPM2 PINs, wiped immediately after the auth value is derived.
- The derived TPM2 auth value and the raw unsealed volume key, wiped after the TPM consumes them / after the base64 form the caller needs is produced.

## Ownership rule 

A passphrase that successfully unlocks a volume is appended to `passphraseCache` **by reference** so later volumes (e.g. btrfs RAID1 members) unlock without re-prompting. That slice is thereafter owned by the cache and is scrubbed exactly once, at the boot exits. Only unowned secrets — failed attempts and consumed PINs — are wiped at their read site. This invariant is documented on the cache and is what keeps the shared-passphrase RAID1 path intact.

## Out of scope

Two secret copies remain because they are handed to an external API as a Go `string`, which is immutable and cannot be zeroed in place.

- TPM2 auth value, the residual string copy. `tpm2Unseal` uses the legacy `go-tpm/legacy/tpm2` API, whose auth argument is a `string` (`UnsealWithSession`). A `[]byte` auth requires porting the unseal path onto the newer `go-tpm/tpm2` transport API which is a possible bigger future project.
- FIDO2 PIN. `go-libfido2` accepts the PIN only as a `string` and copies it to the C heap via `C.CString`; it exposes no `[]byte` variant, and libfido2 itself buffers the PIN internally. Scrubbing it would need upstream changes to go-libfido2 and libfido2. 

## Testing

- New unit tests (`init/secret_test.go`) cover `wipe`, `wipeSecretCache` (incl. idempotency), the failed-attempt wipe, the cache-retention-on-success path, and the TPM2 PIN read/wipe.
- QEMU integration suite, unlock-relevant groups: LUKS1/2 (name/UUID/detached-header), all `Crypttab*` including `CrypttabTPM2` (swtpm), Clevis Tang + Tpm2, `PassphraseCache`, and `LuksBtrfsRaid1` — 27/27 pass. `PassphraseCache` and `LuksBtrfsRaid1` specifically exercise the by-reference cache path the ownership rule protects.

### DIFF
--- a/init/luks.go
+++ b/init/luks.go
@@ -1013,6 +1013,22 @@ func unmarshalTPM2Field(raw json.RawMessage) (value string, sharded bool, err er
 	return s, false, nil
 }
 
+// readTPM2PINAuthValue prompts for a TPM2 PIN, derives the systemd-compatible
+// auth value, and wipes the PIN bytes before returning. An empty PIN yields
+// errTPM2Skipped so the caller falls through to the passphrase prompt.
+func readTPM2PINAuthValue(ctx context.Context, prompt string, salt []byte) ([]byte, error) {
+	pin, err := askKeyboardPassword(ctx, prompt, "")
+	if err != nil {
+		return nil, err
+	}
+	if len(pin) == 0 {
+		return nil, errTPM2Skipped
+	}
+	authValue := tpm2PINAuthValue(pin, salt)
+	wipe(pin)
+	return authValue, nil
+}
+
 func recoverSystemdTPM2Password(ctx context.Context, t luks.Token, mappingName string, tpm2Signature string) ([]byte, error) {
 	var node struct {
 		Blob       json.RawMessage `json:"tpm2-blob"` // base64 string, or array of strings when sharded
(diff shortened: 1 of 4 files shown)


anatol:
@pilotstew thank you! the idea makes a lot of sense.

There is a standard lib feature to do wipe as well https://pkg.go.dev/runtime/secret - have you considered it?

pilotstew:
I looked at it briefly. It's something we could consider. The work here is mostly complementary to it. I didn't go that route because I didn't want to push the build to experimental. When it's stable it should allow better handling of string wipes.

pilotstew:
It's probably the best will do in regards to FIDO wipes. It still wouldn't empty the C heap, which should get dumped later in boot...I'd like it tighter but that's an upstream change. 

TPM can be done. It's another big project but migrating further off the legacy branch is something to tackle when I have more time.

anatol:
Got it. The changes look good. Thank you for taking care of it!

thePrivacyFanatic:
That's nice, it can also can be cool in the future, when such an option is easily available to (along with also covering the things currently out-of-scope) switch from zeroing to shredding (overriding multiple times with random data) to resist such attacks even more

PR #406, opened 2026-07-30
huge manpage cleanup

### LINKED ISSUE #405: inconssitency in manpage
So i'm currently working on improving the manpage (PR coming soon) and two parts look like they contradict.
the crypttab part specifies LUKS options with specific handeling as `keyfile`,`header`,`fido2-device`,`keyfile/token-timeout`,`tpm2-measure-pcr` and `tpm2-signature` while the rd.luks.options section in boot time kernel parameters specifies  `discard`, `same-cpu-crypt`, `submit-from-crypt-cpus`, `no-read-workqueue`, `no-write-workqueue`, `token-timeout`, `tpm2-measure-pcr` and `tpm2-signature`.
are both applicable in both or is there some restriction? (except for the trivial header only being auto-included via crypttab)

thePrivacyFanatic:
firstly note that i accidentally said this was the README in #405 now corrected.
so I've gone over the whole manpage trying to remove any duplications (shrunk by close to 80 lines), specify language in code blocks and to create the links to sections where references were used before.
note that in parts that I've never used (e.g. ssh, FIDO, tang) my understanding was limited so I couldn't verify they were correct (same for examples) so I only rephrased and removed duplications.
even after going over the finished thing I'm not 100% sure I haven't missed something so I'm open to other things that could also use some improving.

note for the machine: 
closes #405

### DIFF
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -171,7 +171,7 @@ Some parts of booster boot functionality can be modified with kernel boot parame
  
  * `rd.luks.data=$UUID=$deviceref` data device for the LUKS volume with UUID as the device itself does not hold it.
  
- * `rd.luks.options=opt1,opt2` supports `discard`, `same-cpu-crypt`, `submit-from-crypt-cpus`, `no-read-workqueue`, `no-write-workqueue`, `tpm2-measure-pcr=no`, `tpm2-signature=<path>` with a higher priority than [crypttab](#crypttab) and persistent flags in header. (note that this is universal but a device-specific option is planned which will also not fail boot when an unknown option is inserted). 
+ * `rd.luks.options=opt1,opt2` supports `discard`, `same-cpu-crypt`, `submit-from-crypt-cpus`, `no-read-workqueue`, `no-write-workqueue`, `tpm2-measure-pcr=no`, `tpm2-signature=<path>` with a higher priority than [crypttab](#crypttab) and persistent flags in header. (Note that this is universal but a device-specific option is planned which will also not fail boot when an unknown option is inserted). 
  
  * `rd.modules_force_load` a comma-separated list of extra kernel modules which should be force loaded.
  
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -375,9 +375,9 @@ Distributions that already drive kernel installation through `kernel-install` (e
 > The legacy Arch-only helper `/usr/lib/booster/regenerate_uki` is deprecated: it does not embed the kernel command line and is therefore vulnerable to trivial attacks.
 
 ### TPM2 auto-unlock and supplantation defense
-secure-boot+disk encryption setups auto-unlock the volume, using a TPM2 enrolled key that is only unsealable when a signed image is booted. but a naive setup (PCR#7 only) is suceptible to the insertion of a malicious unlockable volume with a system that steals the key.
+For setups which aim to protect against hardware access, secure-boot+disk encryption setups enssure the volume key is only unsealable when a signed image is booted. but a naive setup (PCR#7 only) is suceptible to the insertion of a [malicious unlockable volume](https://oddlama.org/blog/bypassing-disk-encryption-with-tpm2-unlock/) with a system that steals the key.
 
-The two methods of countering this are to prevent access to the key after initramfs exit and to bind the initramfs to the exact volume (e.g. by embedding the LUKS header into the image).
+This is prevented by either preventing access to the key after initramfs exit or by binding the initramfs to the exact volume (e.g. by embedding the LUKS header into the image).
 
 Preventing post-initramfs access to the TPM key is done by either binding to a signed PCR#11 policy for the values before the initrd finishes (upon which it measures `leave-initrd` into the PCR) or by binding to an empty value of PCR#15 (which gets all LUKS keys measured into it).  
 both versions prevent access to the key after exiting the initramfs, and additional volumes can be added easily via keyfiles on root and crypttab.
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -7,7 +7,7 @@ booster(1) -- fast and secure initramfs generator
 Booster advantages:
 
  * Fast image build time and fast boot time.
- * Out-of-box support for LUKS-based full disk encryption setup.
+ * Out-of-box support for LUKS-based full disk encryption setups.
  * Clevis style data binding. The encrypted filesystem can be bound to a TPM2 chip or to a network service. This helps to unlock the drive automatically but only if the TPM2/network service is present.
  * Automatically detects and unlocks systemd-cryptenroll (fido2 and tpm2) types of partition encryption.
  * Easy to configure.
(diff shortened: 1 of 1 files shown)


anatol:
cc @pilotstew

pilotstew:
Love the new anchors and de-dupe work. I'd like to propose some collaborative edits I noticed on first scan but I'll need some time to go through it again. Love the direction overall, but I noticed a few spelling errors and some of the areas... UKI, FIDO, TPM need a closer look.

docs/manpage.md
```diff
@@ -148,41 +148,29 @@ Unpack image. Usage: `booster [OPTIONS] unpack image output-dir`
...
- * `rd.luks.options=opt1,opt2` a comma-separated list of LUKS flags. Supported options are `discard`, `same-cpu-crypt`, `submit-from-crypt-cpus`, `no-read-workqueue`, `no-write-workqueue`. `token-timeout=<duration>` sets how long to wait for hardware tokens (FIDO2, TPM2) before also prompting for a keyboard passphrase. Accepts a decimal number followed by a unit (`s`, `m`, `h`), or a bare integer treated as seconds. Default is 30 s. `tpm2-measure-pcr=yes` or `tpm2-measure-pcr=no` forces or suppresses the PCR 15 re-unseal latch for the device (see **TPM2 auto-unlock and supplantation defense**); the command-line value takes precedence over crypttab. `tpm2-signature=<path>` (or `false`) selects or disables the PCR signature file for a signed (authorized) policy enrollment.
-    Note that booster also supports LUKS v2 persistent flags stored with the partition metadata. Any command-line options are added on top of the persistent flags.
+ * `rd.luks.header=$UUID=$PATH` detached LUKS header for LUKS volume with UUID.`$PATH` can be an absolute path in the initramfs (including block devices) or a `$path:$deviceref` which will temporarely mount the device as read-only.
```
pilotstew: Might be worth mentioning to get the header into the initramfs it needs to be included in extra_files:

docs/manpage.md
```diff
@@ -148,41 +148,29 @@ Unpack image. Usage: `booster [OPTIONS] unpack image output-dir`
...
+ * `rd.luks.data=$UUID=$deviceref` data device for the LUKS volume with UUID as the device itself does not hold it.
+ * `rd.luks.options=opt1,opt2` same options but higher priority than [crypttab](#crypttab) options column except for the lack ov build-time proccessing. note that this is universal but a device-specific option is planned.
```
pilotstew: When booster encounters an unknown rd.luks.options it hard-errors and dumps to emergency shell if enabled. The previously listed flags are as far as I know the only ones accepted. This reads as if any crypttab option is viable here.

typos:
ov -> of
note -> Note

docs/manpage.md
```diff
@@ -192,44 +180,37 @@ on one of the configurations below.
...
-architectures use different shortcuts; consult your editor's type list.
-The full per-architecture GUID list is in the [Discoverable Partitions Specification](https://uapi-group.org/specifications/specs/discoverable_partitions_specification/).
+Set the root partition's GPT type to the [Discoverable Partitions Specification](https://uapi-group.org/specifications/specs/discoverable_partitions_specification/)'s "Linux root" for your CPU architecture
+using any GPT partition editor (gdisk, sgdisk, fdisk, cfdisk, parted, ...). consult your editor's type list for specifics.
 
-The same GUID covers both plain-filesystem and LUKS-encrypted roots —
-booster handles either. Booster scans only the disk that holds the
-active EFI System Partition; root-tagged partitions on other disks are
-ignored. Tag exactly one partition on that disk.
+The same GUID covers both plain-filesystem and LUKS-encrypted roots. Booster scans only the disk that holds the active EFI System Partition.
```
pilotstew, line 235: If multiple partitions on the same device are labeled as root, booster will automatically select the first in number order. This was the reason for the "Tag exactly one partition" comment. Which was poorly worded but may still be relevant because there are no warnings and the result may be unexpected.

docs/manpage.md
```diff
@@ -369,95 +335,52 @@ Device reference has one of the following values:
...
+> when using plymouth the new builds should automatically clear the prompt right upon unlock.
 
-**Cancel-on-win.** Any prompt — keyboard passphrase, FIDO2-PIN, or TPM2-PIN — is dismissed automatically when a parallel unlock path wins (a touchless token completes, a sibling volume's keyfile reads successfully, etc.). This applies on both the kernel console and the Plymouth splash. On Plymouth the on-screen prompt clears immediately on the server side once Plymouth MR !393 is picked up; older Plymouth builds release the prompt server-side but leave its UI drawn until the splash is otherwise cleared.
+For LUKS volumes with multiple FIDO2 tokens enrolled booster checks each device for the credentials before prompting for PIN to skip irrelevant devices.
```
pilotstew: This was previously worded incorrectly.. Before prompting for a PIN, booster asks each attached FIDO2 authenticator whether it holds this volume's credential. Devices that do not hold it are skipped. This runs whenever a FIDO2 token is enrolled, and does not need multiple enrolled. This allows a machine with several keys attached, or with the wrong key attached to reach the next unlock method without a spurious or blocking prompt. If no FIDO2 device is attached at all, booster instead waits for one to be plugged or token timeout expires. Tokens enrolled with fido2-uv-required=true skip the pre-flight and go straight to the full assertion....This was a v13 addition allowing a user who had a primary and secondary key not get blocked by the prompt for a missing primary key. OR in the strange event they had both inserted.

That said, brevity is good. But I would rephrase slightly for clarity here. Something like:

Booster checks FIDO devices and only prompts for PIN if the devices credentials match an enrolled token. If no devices match an enrolled token, booster will wait until token timeout or the user advances the prompt.

Open to changes on that wording.

docs/manpage.md
```diff
@@ -369,95 +335,52 @@ Device reference has one of the following values:
...
-
-Star symbol `*` is a shortcut for "all modules", it can be used to add all modules or remove all modules from the image.
-
-Next booster moves to the `modules_force_load` option that consists of module names to load at the boot time.
-All these modules are also added to the image.
-
-At the final step booster computes dependency graphs between modules and all required dependencies.
-For example if a user manually added `ext4` and kernel build system says `ext` module requires `mbcache` and `jbd2` then both
-`mbcache` and `jbd2` automatically added to the image.
+Booster has a list of that are likely to be needed for boot hardcoded (as `defaultModuleList` in `generator.go`).
```
pilotstew: "list of [modules] that are likely"?

pilotstew:
I need to take another look at the supplantation defense section. The original could definitely be more concise. But as it's written now it sounds like PCR11 OR PCR15 is sufficient which isn't quite correct. I should have something for you tomorrow.

pilotstew:
Becuase the example section descibes the setup, the TPM2 autounlock and defense section can be simplified. I'm think something along these lines. It states the requirement for all 3 PCR's and passes the vulnerability off to oddlama's blog. 

Booster can unseal a LUKS volume key from a TPM2 token and unlock the root unattended. However, an attacker can present a rogue root filesystem that leaves PCR registers unchanged and still be handed the real key (see https://oddlama.org/blog/bypassing-disk-encryption-with-tpm2-unlock/). Booster's recommended binding is therefore a signed PCR 11 measuring the UKI's kernel, initrd and command line together with PCR 7 and an all-zero PCR 15, which booster extends with `HMAC(volume_key, "cryptsetup:" + name + ":" + uuid)` on unlock preventing subsequent unlocks. For additional unlocks use a keyfile on the first device. A backup passphrase is recommended in the event of PCR mismatch. See the Headless TPM2 example under EXAMPLES has the configuration. 

I didn't include your "bind the initramfs to the exact volume (e.g. by embedding the LUKS header into the image)." because this statement is more accurately describes the newer  `fixate-volume-key=` which is not yet supported but is on my ever growing list of todo's. 

What do you think @thePrivacyFanatic?

thePrivacyFanatic:
I agree this part could use some changes but I think our ideas for it were a little different.
my understanding is that your ideal page emphasizes the practical examples while my plan was to first mostly list the tools that it provides and keep setup details to the examples section.
My idea is based on believing that an initramfs generator as a tool is advanced enough that the user can be expected to understand their wanted setup just from knowing what they have in their disposal, but I absolutely may be wrong on that so I'm not sure about the best path foreward.

thoughts?

P.S. about the `fixate-volume-key=` option, I was thinking as long as the header is included (and ofc specified in options) the volume key is locked to the one in it, but the connected header option could be really nice for easier recovery. so just to understand, what is your plan for it to do?

pilotstew:
I think I have a different expectation of users knowledge base and my goal has been to head off issues for misconfiguration. But your framing is helpful and I think correct. We don't need to explain the threat specifically the example given serves as the recommended default tpm Auto Unlock. 

In that light I think it might be fine just to drop the whole section. 

My plan for fixate-volume-key is just to make it available. Adding it to cryptab should be pretty easy but cmdline relies upon the device specific changes. The arch wiki has a small section on its use https://wiki.archlinux.org/title/Dm-crypt/System_configuration

thePrivacyFanatic:
> In that light I think it might be fine just to drop the whole section.

probably just add a part about it in one of the examples or in the notes section
alternatively demote it (and maybe also UKI) to notes instead of full sections

thePrivacyFanatic:
I've elected in the end to have it in notes, just making sure you don't think UKI needs to be categorized as a section there as I've demoted it too 
(also just a little clearing up of things)

docs/manpage.md
```diff
@@ -245,45 +226,30 @@ and rebuild.
...
- * **`tpm2-measure-pcr=`** — `yes` or `no`. Overrides the PCR 15 re-unseal latch (see **TPM2 auto-unlock and supplantation defense**). When unset, booster extends PCR 15 automatically if the device's `systemd-tpm2` token is bound to PCR 15; `yes` forces the extend, `no` suppresses it. The same option on the kernel command line (`rd.luks.options=`) takes precedence.
- * **`tpm2-signature=`** — path to a systemd PCR signature JSON for a signed (authorized) PCR policy (see **TPM2 auto-unlock and supplantation defense**); `false` disables signed-policy unlock. When unset, booster reads the signature that systemd-stub unpacks from the UKI into the initramfs at `/.extra/tpm2-pcr-signature.json`. **A non-UKI boot populates no such file**, so there you must supply it explicitly with `tpm2-signature=<path>` (generate it with `systemd-measure sign`) or signed-policy unlock falls through to the passphrase. Also accepted on the kernel command line via `rd.luks.options=`.
+ * **`keyfile=`**: path of form `/path:deviceref` to keyfile on a separate device. Booster temporarily mounts the device read-only at boot to read the key.
```
pilotstew: The placement of this is misleading. key-file declaration is in the password column of /etc/crypttab. The distinction was always too subtle with the missing "="  but it's actually field #3 (key-file declared password) not a crypttab option=value token.

That said this line parrots the crypttab spec and can therefore be removed entirely, redundant.

pilotstew (commented):
I apologize these were my initial comments on some of the relevant sections, I thought I submitted it long ago.

docs/manpage.md
thePrivacyFanatic: you're right. I just tested redonly and it fails the boot

docs/manpage.md
thePrivacyFanatic: since it's AFAIK undefined behavior in other places i have opted to put only a reminder there but this makes a lot of sense.

docs/manpage.md
thePrivacyFanatic: to have this ordered i think I'd split FIDO2 into its own section as it has more relevant info than should be in the prompt order one

docs/manpage.md
thePrivacyFanatic: yeah, haven't noticed, thanks.

docs/manpage.md
thePrivacyFanatic: is that not an option in rd.luks.options? if it is I'll have to maybe separate those into a LUKS options section

docs/manpage.md
pilotstew: I'm nearly done with the initial rd.luks.options=UUID PR. It will resolve much of these issues. Unknown options will warn rathar than fail. Several unimplemented options available currently only via crypttab will follow the initial PR. As well as some new ones. If you want we can just hold rd.luks.options=opt1,opt2 as is. That PR can probably be ready today.

docs/manpage.md
pilotstew: Yeah, makes sense.

docs/manpage.md
pilotstew: kernel cmdline:  rd.luks.key=
crypttab:
 ```
 <name>       <device>             <password>               <options>
 root        /dev/sda          /keyfile or password         
 ````
 
systemd's crypttab(5) defines only the following modifiers as options: keyfile-erase, keyfile-offset=, keyfile-size=, keyfile-timeout=

keyfile-erase is on my list but not supported yet...Same vein as my recent zeroing work.

pilotstew: In some sense because this is already well documented in crypttab(5) and functions as expected in booster it may not be worth keeping. Save explanation for items where booster's handling differs.

thePrivacyFanatic: makes sense

pilotstew:
Changes look good to me

thePrivacyFanatic:
also think this is ready

anatol:
@thePrivacyFanatic there is currently merge conflict. could you please resolve it and I land the PR to the repo.

thePrivacyFanatic:
done

anatol (approved):

PR #407, opened 2026-08-06
init: scan every btrfs member as it is discovered

### LINKED ISSUE #283: Failed to mount btrfs array
Related to #194.
`booster.yaml` is empty.
Kernel command line: `rd.luks.name=7675b74a-f03d-4aa5-ae60-b687a432a426=root1 rd.luks.name=e1400000-62b5-47cd-bfd1-2795ea331ace=root2 root=/dev/mapper/root1 rootflags=subvol=@`

When booting, after typing the first password, it immediately fails. I tried typing the password for the second drive afterwards, it says `Unlocking...` but then nothing happens.
```
Enter passphrase for root1: Unlocking...
Enter passphrase for root2: [   8.96973] BTRFS error (device dm-0): devid 2 uuid 885... is missing
[    8.96978] BTRFS error (device dm-0): failed to read the system array: -2
[    8.96987] BTRFS error (device dm-0): open_ctree failed: -2
[    8.96994] booster: mount(/dev/dm-0): no such file or directory
  Unlocking...
```

Related to that, when using mkinitcpio with the sd-encrypt hook, the same password is tried for subsequent drives, avoiding have to type the same password multiple times.

pilotstew:
Closes #283 

A multi-device btrfs root fails to boot when `root=` names a mapper device rather than the filesystem UUID. The passphrases are accepted, the volume never assembles, and the boot sits there until the ten-minute btrfs readiness timeout expires. With `loglevel=6` there is nothing on the console after the last prompt, because the once-a-second progress line is logged at KERN_INFO and the kernel suppresses it — so it looks like a silent hang.

The cause is that booster only ever registers a btrfs device with the kernel as a side effect of preparing to mount it. `waitForBtrfsDevicesReady` issues `BTRFS_IOC_DEVICES_READY`, which scans the one device it is asked about, and it is only reached for a device matching `root=`. A volume is ready only once every member has been scanned.

Mounting by filesystem UUID hides this completely: every member of a btrfs volume carries the same fs UUID, so `root=UUID=...` matches all of them and each gets scanned on its way into the readiness wait. `root=/dev/mapper/<name>` matches exactly one, and its siblings are never registered.

This registers each btrfs member with `BTRFS_IOC_SCAN_DEV` as it is discovered, which is what `btrfs device scan` and udev's btrfs rules do on a normal boot. A scan failure is logged and the boot continues rather than aborting. The ioctl argument struct, the magic and the opcodes move out of `waitForBtrfsDevicesReady` so both call sites share them, along with the control-node open — which also loads the btrfs module, since `/dev/btrfs-control` only exists once it is in.

`TestLuksBtrfsRaid1MapperRoot` covers it: a btrfs RAID1 over two LUKS2 members with different passphrases, unlocked via `rd.luks.name=` and mounted via `root=/dev/mapper/root1`. Without the fix it hangs until the VM times out; with it the volume assembles and the boot completes. Both members are prompted exactly once, in either order, since device discovery order is not deterministic.


anatol:
Thank you, it looks good. Merging it now.

Given that the amount of btrfs-specific code is getting bigger in main.go it probably worth splitting all that logic to a separate btrfs.go file.

PR #408, opened 2026-08-06
init: move the btrfs device handling into its own file

pilotstew:
The btrfs-specific code in main.go had grown to the ioctl argument struct, the opcodes, the control-node open, the scan and the readiness wait.  Move all of it to btrfs.go.

Pure movement: no logic changed, and the dispatch that calls into it stays in processBlkInfo and mountRootFs alongside the other filesystem handling.


anatol:
Thank you!

Issue #403, opened 2026-07-18
[BUG] USB-Key Problem (Booster doesn't ask for a password whenever key is removed)


glxguard:
## Current Behavior

Hello! I use 3 LUKS-encrypted disks. I have two keys to unlock them: a password, and a 4kb keyfile I have.

I'll better explain this way:

2Gb USB Flash (EROFS filesystem) ─> Has a keyfile named .key

```
.key             PASSPHRASE
   │                  │ 
   ├── root partition ┤
   ├─ home partition ─┤
   └── hdd partition ─┘
```

**How it shoud work:**
USB Plugged ─> I do not need a passphrase
USB Unplugged ─> Booster asks for password

**How it really does:**
USB Plugged ─> I do not need a passphrase
USB Unplugged ─> Booster is stuck unable to unlock root parition

## About my configuration

Here are my **boot options**:

`rd.luks.name=[ROOT UUID]=root   root=/dev/mapper/root   rw   initrd=\booster-linux.img`

Here's my **crypttab**:

```
# Diskname | Identifier | Keyfile | Options
root  UUID=[ROOT UUID]  /.key:UUID=[USB UUID] x-initrd.attach
home UUID=[HOME UUID] /.key:UUID=[USB UUID] x-initrd.attach
hdd UUID=[HDD UUID] /.key:UUID=[USB UUID] x-initrd.attach
```

## Additional information

- Booster version: 0.13-1 
- OS: Artix Linux with dinit

pilotstew:
Set keyfile-timeout in your crypttab options. It defaults to `keyfile-timeout=0` which means, wait infinite. So when you unplug the keyfile USB it waits forever.

glxguard:
@pilotstew Thanks! That helped!

PR #409, opened 2026-08-06
Pr/keyfile timeout default

pilotstew:
When neither keyfile-timeout= nor mount_timeout is set, booster waits indefinitely for the device holding a keyfile. If that device never appears, the boot stops there and the passphrase prompt is never reached.

This gives keyfile-timeout  the same behavior as token-timeout setting the default to 30s, 0s means wait forever. See issue #403


anatol:
cc @thePrivacyFanatic

thePrivacyFanatic:
explanation makes full sense

Issue #410, opened 2026-08-07
Won't work with secureboot on


RoninByte:
Hi,

I switched from dracut to booster on fedora and it works until I turn on secure boot when I turn it on, my laptop just gets stuck at the laptop boot logo and pressing escape just put a flashing line in the top left corner I'm not sure why this is happening, on dracut it booted fine with secureboot on, it has been signed with sbctl and I use systemd-ukify

pilotstew:
Can you post the result of `sbctl verify`  and `sbctl status`

Also your `/etc/kernel/uki.conf` it should look something like:

```
[UKI]
Microcode=/boot/amd-ucode.img            # or intel-ucode.img
SecureBootPrivateKey=/var/lib/sbctl/keys/db/db.key
SecureBootCertificate=/var/lib/sbctl/keys/db/db.pem
```

Fedora automatically updates the UKI on system updates via kernel-install if you signed them manually but not via the config file the UKI may have been overwritten and without the signing SercureBoot denies the image.

Check if you have `/usr/lib/kernel/install.d/91-sbctl.install` plugin enabled that signs image on update. If configured correctly uki.conf shouldn't actually need the SecureBoot keys or cert.

RoninByte:
> Can you post the result of `sbctl verify` and `sbctl status`
> 
> Also your `/etc/kernel/uki.conf` it should look something like:
> 
> ```
> [UKI]
> Microcode=/boot/amd-ucode.img            # or intel-ucode.img
> SecureBootPrivateKey=/var/lib/sbctl/keys/db/db.key
> SecureBootCertificate=/var/lib/sbctl/keys/db/db.pem
> ```
> 
> Fedora automatically updates the UKI on system updates via kernel-install if you signed them manually but not via the config file the UKI may have been overwritten and without the signing SercureBoot denies the image.
> 
> Check if you have `/usr/lib/kernel/install.d/91-sbctl.install` plugin enabled that signs image on update. If configured correctly uki.conf shouldn't actually need the SecureBoot keys or cert.

` sudo sbctl verify
Verifying file database and EFI images in /boot/efi...
✓ /boot/efi/EFI/Linux/fedora-7.1.5-201.fc44.x86_64.efi is signed
✓ /usr/lib/systemd/boot/efi/systemd-bootx64.efi.signed is signed
✓ /boot/efi/EFI/BOOT/BOOTX64.EFI is signed
[...]

PR #411, opened 2026-08-10
fedora packaging

pilotstew:
packaging/centos cannot be built on a current Fedora...spec fails to parse, its BuildRequires name CentOS 7 yum plugins, and %files omits 60-booster.install.

Adds packaging/fedora/booster.spec and a build helper. File coverage matches the Arch package. The yum transaction hooks are dropped, since kernel-install handles regeneration.

See #410


anatol:
Look good to me. Thank you!

PR #412, opened 2026-08-14
Add systemd's per-device rd.luks.options=$UUID= and compose the option sources at one point

pilotstew:
Booster takes `rd.luks.options=` as a global list only. Options that describe a single volume have no per-device spelling on the kernel command line: a detached header, a key slot, a retry count, a keyfile timeout. The only place to set them per device is `/etc/crypttab`, which is baked into the image at build time and cannot be countermanded from the bootloader. This adds systemd's `rd.luks.options=$UUID=opt1,opt2` and puts the option sources under one composition step with a stated order.

## What's new

- `rd.luks.options=$UUID=opt1,opt2` scopes options to one device and takes the fourth crypttab field's whole vocabulary, rather than the subset the global list understood.
- A per-device list *replaces* that device's crypttab options instead of adding to them, so the command line can remove an option and not only add one. That is what makes it usable to boot past a wrong entry. `systemd-cryptsetup-generator(8)` says the per-UUID form "will override the options for the given UUID", and calls the parameter the analogue of the fourth crypttab field. The entry's first three fields are unaffected: the device, the name and the key file survive. What was dropped is named in the boot log.
- `rd.luks.header=` is deprecated in favour of `rd.luks.options=$UUID=header=`, the systemd spelling for the same thing. It still works and warns at boot. Where both name the same device, the option form wins.

## One place, one order

Options used to be applied as they were parsed. The UUID-less `rd.luks.options=` list was pushed onto every mapping at the end of `parseParams`. A key file given without a UUID was attached the instant that parameter was read. Crypttab was merged in later, during `boost()`. No step could see the others' work, so which source won followed from which ran first rather than from any stated rule, and a mapping that did not yet exist when a step ran never received what it applied.

Every source is now held until all of them are known, then composed once, in one place:

`/etc/crypttab` → `rd.luks.options=` → `rd.luks.header=` → `rd.luks.options=$UUID=`

Later overrides earlier, option by option. dm-crypt flags accumulate across sources instead of replacing. A per-device list is the one exception to "option by option": it replaces that device's crypttab options whole. The precedence rule is the order of those four lines, so changing it means reordering them rather than adjusting conditions spread across three functions. The same rule covers every device, however it was declared: named on the command line, declared only in `/etc/crypttab`, or found by autodiscovery.

No parameter's meaning now depends on what was read before it. A test resolves one set of parameters in every rotation and requires a single outcome.

## Behaviour changes

- An unrecognised option no longer aborts the boot. On the command line it was fatal. In crypttab it was debug-logged and invisible without `booster.log=debug`. Both now warn and skip, as systemd-cryptsetup does. A malformed value for a recognised option stays fatal.
- A key file given without a UUID is a default for every device the command line does not give one of its own, which is what systemd documents. It used to demand that exactly one mapping existed at the moment it was parsed and abort the boot otherwise, so it failed for two or more devices, for none, and whenever it was written before the parameter naming the device.
- An entry booster will not unlock is reported as discarded, rather than emitting a warning per option on a line it was never going to act on. That covers `noauto` and the volume types booster does not support.

Two long-standing defects surfaced while restructuring: an explicit `tries=0` overwritten by a crypttab `tries=`, and a dm-crypt flag passed to the kernel twice. Both are fixed in the commits that touch that code. Smaller ones introduced and resolved within this series are described in their own commit messages.

## Testing

- New boot tests cover the per-device form replacing a crypttab entry's options, a detached header configured entirely through the option form, and the command line's options surviving a crypttab that fails to parse.


anatol:
cc @thePrivacyFanatic for more attention

docs/manpage.md
```diff
@@ -260,9 +262,18 @@ the file by either being ran as root, or by being passed
...
 
 
-Crypttab entries and `rd.luks.*` parameters with the same LUKS UUID are 
-merged like in systemd-cryptsetup. with cmdline parameters taking
-precedence and unspecified options being added from crypttab.
+Crypttab entries and `rd.luks.*` parameters with the same LUKS UUID are
+merged, with cmdline parameters taking precedence and unspecified options
+being added from crypttab.
+
+A per-device `rd.luks.options=$UUID=` is the exception: it *replaces* that
```
thePrivacyFanatic: (forgive me for not marking the whole paragraph, firefox was causing trouble)
I do think this can be shortened a little
maybe something like
```A per-device `rd.luks.options=$UUID=` is an exception, *replacing* the
entry's options to allow for excluding universal options from said entry. 
(this matches systemd-cryptsetup.)```

File: docs/manpage.md [hunk 2]
```diff
@@ -276,6 +287,8 @@ Booster-specific behaviour for selected options:
  
  * **`tpm2-signature=`**: path to a systemd PCR signature JSON for a signed (authorized) PCR policy. `false` disables signed-policy unlock. When unset on UKI boot booster reads the signature that systemd-stub unpacks from the UKI into the initramfs at `/.extra/tpm2-pcr-signature.json`. otherwise required for signed-policy unlock.
 
+ * **unrecognised options**: reported and skipped, not fatal. Booster implements a subset of crypttab(5), so options it does not act on (`headless=`, `keyfile-erase`, `fixate-volume-key=`, …) are named in the boot log. `_netdev` is accepted — there is no unit ordering to apply it to — but warns when the entry needs the network and none is configured. An entry for a volume type booster cannot unlock (`swap`, `tmp`, `plain`, `bitlk`, `tcrypt`) is skipped whole and reported as such, without complaining about its options.
```
thePrivacyFanatic: I think it would be more noticeable as a normal paragraph, and can be shortened to something similar to
```Unrecognized options are reported and skipped.``` if the other parts about error handling are less important, or ```Unrecognized options are reported and skipped, while _netdev warns if no network is configured and entries for unsupported volume types are skipped and reported.``` of they do matter here

anatol:
The feature looks good

anatol:
Thank you @pilotstew !

Issue #401, opened 2026-07-14
Major regression in binary size between 0.12 and 0.13


cyyynthia:
It appears the `init` binary has almost doubled in size between Booster 0.12 and Booster 0.13, which has an impact of 4MB on the final binary. For setups without FDE nor LVM that's double the size of the final binary, and ~30% larger for setups with FDE and LVM.

While I guess it's expected as Booster added a lot of new features in the 0.13 release, Booster is larger than `mkinitcpio` with just the `base` hook and NVMe modules, and `mkinitcpio`'s very fast with just that hook too. With FDE, Booster is much faster than `mkinitcpio` and remains smaller, but only by 4MB (and without advanced systemd's features AFAICT; plus I couldn't unlock my drive with my Yubikey). IOW, it's 4MB smaller but with less mature code and more room for failure)

Note that I use a custom script to extract the generated image; this is because Booster 0.13 broke `unpack`: it can't extract its own archives because of an issue with symlinks (on my system).

```
# pacman -S booster
# cat /etc/booster.yaml /etc/crypttab
compression: zstd
extra_files: fido2-assert
vconsole: true
enable_fido2: true
enable_lvm: true
# Configuration for encrypted block devices.
# See crypttab(5) for details.
crypt0    	UUID=333d089e-d99b-46e5-bf77-c4e2124c12f6	none      	tpm2-measure-pcr=yes,tpm2-device=auto,fido2-device=auto,x-initrd.attach
# booster build test.img -f; du test.img
16508	test.img
# rm -rf a; booster ls test.img | while read f; do if [[ ! ( "$f" =~ "->" )]]; then mkdir -p "a/"`dirname $f`; booster cat test.img "$f" > "a/$f"; fi; done
# find a -type f -exec du -a {} + | sort -n -r
<See `Booster 0.13-1 (FDE)` below>
# pacman -U https://archive.archlinux.org/packages/c/booster/booster-0.12-2-x86_64.pkg.tar.zst
# booster build test.img -f; du test.img
12444	test.img
# rm -rf a; booster ls test.img | while read f; do if [[ ! ( "$f" =~ "->" )]]; then mkdir -p "a/"`dirname $f`; booster cat test.img "$f" > "a/$f"; fi; done
# find a -type f -exec du -a {} + | sort -n -r
<See `Booster 0.12-2 (FDE) below>
```

Without FDE:

```
# pacman -S booster
# cat /etc/booster.yaml /etc/crypttab
compression: zstd
[...]

PR #414, opened 2026-08-19
Reduce init binary and image size by modularizing Clevis into a plugin.

mikispag:
This PR reduces the size of the booster `init` binary and the generated initramfs image by modularizing Clevis / Tang network unlock into a dynamic plugin (partially addresses #401).

1. **Clevis Plugin (`clevisplugin.so`)**:
   - Extracted `clevis.go` and its transitive dependencies (`lestrrat-go/jwx`, `net/http`, `crypto/tls`, `crypto/x509`) into `init/clevisplugin.so` following the existing `fido2plugin` pattern.
   - Core `/init` dynamically loads `clevisplugin.so` via `plugin.Open()` when a Clevis token is encountered.
   - The generator bundles `clevisplugin.so` only when `enable_clevis: true` is configured.

### Size Impact

| Artifact | Before | After | Reduction |
| :--- | :--- | :--- | :--- |
| **`/init` binary** | `24.0 MB` | **`12.0 MB`** | **-50% (-12 MB)** |
| **Default initramfs image** (zstd compressed) | `22.0 MB` | **`14.0 MB`** | **-36% (-8 MB)** |

### DIFF
--- a/generator/generator.go
+++ b/generator/generator.go
@@ -37,6 +37,7 @@ type generatorConfig struct {
 	readDeviceAliases       func() (set, error)
 	readHostModules         func(kernelVer string) (set, error)
 	readModprobeOptions     func() (map[string]string, error)
+	readHostClevisTokens    func() (bool, error)
 	stripBinaries           bool
 	enableLVM               bool
 	enableMdraid            bool
@@ -56,8 +57,10 @@ type generatorConfig struct {
 	enableVirtualConsole     bool
 	vconsolePath, localePath string
 
-	crypttabFile string // path to host crypttab; empty = use /etc/crypttab
-	enableFido2  bool
+	crypttabFile         string // path to host crypttab; empty = use /etc/crypttab
+	enableFido2          bool
+	enableClevis         bool
+	explicitEnableClevis bool
 
 	serializeTokens bool // dispatch LUKS tokens serially instead of concurrently; default false
 	tokenTimeout    int  // device-level keyboard-fallback timer (seconds); 0 = unset
@@ -137,7 +140,7 @@ func generateInitRamfs(conf *generatorConfig) error {
 	if !explicitCrypttab {
 		crypttabPath = "/etc/crypttab"
 	}
-	if hasFido2, err := img.appendCrypttab(crypttabPath); err != nil {
+	if hasFido2, hasClevis, err := img.appendCrypttab(crypttabPath); err != nil {
 		switch {
 		case explicitCrypttab:
 			return err
@@ -148,9 +151,20 @@ func generateInitRamfs(conf *generatorConfig) error {
 		default:
 			return err
 		}
-	} else if hasFido2 {
-		// auto-enable fido2 plugin when any crypttab entry uses fido2-device=
-		conf.enableFido2 = true
+	} else {
+		if hasFido2 {
+			// auto-enable fido2 plugin when any crypttab entry uses fido2-device=
+			conf.enableFido2 = true
+		}
+		if hasClevis && !conf.explicitEnableClevis {
+			conf.enableClevis = true
+		}
+	}
+
+	if !conf.explicitEnableClevis && !conf.enableClevis && conf.readHostClevisTokens != nil {
+		if detected, err := conf.readHostClevisTokens(); err == nil && detected {
+			conf.enableClevis = true
+		}
 	}
 
 	for _, f := range conf.extraFiles {
@@ -169,6 +183,22 @@ func generateInitRamfs(conf *generatorConfig) error {
 		}
 		if err := img.AppendContent("/usr/lib/booster/fido2plugin.so", 0o755, content); err != nil {
 			return fmt.Errorf("fido2 plugin: %v", err)
+		}
+	}
+
+	if conf.enableClevis {
+		pluginPath := filepath.Join(filepath.Dir(conf.initBinary), "clevisplugin.so")
+		content, err := os.ReadFile(pluginPath)
+		if err != nil {
+			if conf.explicitEnableClevis {
+				return fmt.Errorf("clevis plugin %s: %v", pluginPath, err)
+			}
+			warning("clevis plugin %s unavailable, building without clevis: %v", pluginPath, err)
+			conf.enableClevis = false
+		} else {
+			if err := img.AppendContent("/usr/lib/booster/clevisplugin.so", 0o755, content); err != nil {
+				return fmt.Errorf("clevis plugin: %v", err)
+			}
 		}
 	}
 
@@ -512,6 +542,7 @@ func (img *Image) appendInitConfig(conf *generatorConfig, kmod *Kmod, vconsole *
 	initConfig.EnableZfs = conf.enableZfs
 	initConfig.ZfsImportParams = conf.zfsImportParams
 	initConfig.EnablePlymouth = conf.enablePlymouth
+	initConfig.EnableClevis = conf.enableClevis
 	initConfig.SerializeTokens = conf.serializeTokens
 	initConfig.TokenTimeout = conf.tokenTimeout
 	initConfig.PinDelay = conf.pinDelay
--- a/init/clevis_cgo.go
+++ b/init/clevis_cgo.go
@@ -1,0 +1,31 @@
+//go:build cgo
+
+package main
+
+import (
+	"fmt"
+	"plugin"
+
+	"github.com/anatol/booster/init/clevisiface"
+)
+
+const clevisPluginPath = "/usr/lib/booster/clevisplugin.so"
+
+var clevisPlugin clevisiface.ClevisPlugin
+
+func loadClevisPlugin() error {
+	p, err := plugin.Open(clevisPluginPath)
+	if err != nil {
+		return fmt.Errorf("clevis: cannot open plugin %s: %w", clevisPluginPath, err)
+	}
+	sym, err := p.Lookup("Plugin")
+	if err != nil {
+		return fmt.Errorf("clevis: plugin missing Plugin symbol: %w", err)
+	}
+	pl, ok := sym.(*clevisiface.ClevisPlugin)
+	if !ok {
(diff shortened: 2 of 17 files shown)


init/clevis_cgo.go
```diff
@@ -0,0 +1,47 @@
...
+		if !ok {
+			warning("clevis: Plugin symbol does not implement ClevisPlugin interface")
+			return
+		}
+		clevisplugin = *pl
+	})
+}
+
+func clevisDecrypt(payload []byte) ([]byte, error) {
+	loadClevisPlugin()
```
anatol: i think this plugin should be loaded earlier if `enable_clevis` is set. And fail if plugin cannot be load.

Then the user below just uses `clevis.Decrypt()`. No need for this wrapper function.

@pilotstew what do you think?

mikispag: Agree. See new commit. Also added a `cgo`/`no_cgo` split like you did for `fido2`.

pilotstew:
Love the direction, pulling optional dependencies out of the image has been on my todo list so thanks for taking it on. Clevis, ssh, and internet are the big ones. I measured the generated images, same kernel:

- clevis not enabled: 8.38 MB today, 6.42 MB with this PR
- `enable_clevis: true`: 8.38 MB today, 11.28 MB with this PR

And on a universal image:
- clevis not enabled: 36.30 MB today, 34.34 MB with this PR
- `enable_clevis: true`: 36.30 MB today, 39.20 MB with this PR


The plugin carries its own copy of the Go runtime and the shared dependency set, but clevis users pay for it and it's a reasonable trade.

My biggest concern is the unlike the fido plugin clevis has always been included without a config option. Because of that we must default to `enable_clevis: true` or risk breaking boot for users who fail to update their config file. Unfortunately that means users who don't specify `enable_clevis: false` get a larger img. That said, it's still the right direction and users who care can opt into the size reduction.

As for the packaging flags. It's most likely harmless but at the same time it's value is probably overstated. I think most if not all Arch based packagers default to strip which makes the flag redundant. I tested it on the fedora container and saw no breaking change and the img did shrink but seemingly because init's debuginfo was deleted.
Unclear on centos, but that packaging looks to be stale and still has CGO_ENABLED=0 which would be a problem for fido and this PR. For anyone manually building the flags would absolutely reduce image size.

pilotstew:
FWIW I think the long term solution to these kinds of trades could be luks header read at build time like we do during boot. It would allow for automatic inclusion of plugins but overideable via booster.yaml. A simpler config and less painful on config changes.

mikispag:
Thanks @pilotstew! Ideally, I would like to deliver the size reduction out of the box while preserving full backwards compatibility.

I took your advice and implemented a simple build-time auto-detection:

1. **Backwards Compatible for Existing Clevis Users**:
   - The generator scans host block devices via `luks.Open()` for LUKS2 `"clevis"` tokens and checks `/etc/crypttab` for `clevis`/`tang` options (**warning**: please review this carefully, I am not an expert).
   - If Clevis is in use, `clevisplugin.so` is bundled automatically. Existing Clevis setups will continue booting on upgrade without needing any config changes.
2. **Automatic Size Reduction for Non-Clevis Users**:
   - If no Clevis tokens or crypttab entries exist on the host, the plugin is omitted by default, giving non-Clevis users the smaller image size automatically.
3. **Universal Images**:
   - `universal: true` images default to including the plugin so portable images work on any machine.
4. **User Override**:
   - Explicitly configuring `enable_clevis: true` or `enable_clevis: false` in `/etc/booster.yaml` always takes precedence over auto-detection.

Finally, I tweaked the CentOS packaging target.

pilotstew:
Having thought a bit more on `-ldflags="-s -w"`. I think it's a great thought but the wrong lever. Or more accurately complicates the right levers. These are the options that currently do what you're asking and they do it by default. Hard coding it here would override a user who has explicitly changed that default. Only the build-from-source-user benefits from the change and arguably the right way for them to do this is via go build flags.
```
Source builder, one-off:
    go build -ldflags=-s ./...

Source builder, persistent across future builds:
    go env -w GOFLAGS=-ldflags=-s

Arch packager, already the default (stock makepkg.conf):
    OPTIONS=(strip ...)
    STRIP_BINARIES="--strip-all"

Arch user who wants symbols back:
    OPTIONS=(debug !strip)      # /etc/makepkg.conf
    options=(debug !strip)      # or per-package in the PKGBUILD

Fedora packager, already the default: find-debuginfo strips the binary and ships symbols in booster-debuginfo.

Fedora spec that wants no debuginfo at all:
    %global debug_package %{nil}
```

Also booster has a strip option which I'm getting ready to push changes making it more reliable.

I'll take a look at the new commits.

mikispag:
Squashed into a single, descriptive commit. PTAL.

generator/generator.go
```diff
@@ -172,6 +186,17 @@ func generateInitRamfs(conf *generatorConfig) error {
 		}
 	}
 
+	if conf.enableClevis {
+		pluginPath := filepath.Join(filepath.Dir(conf.initBinary), "clevisplugin.so")
+		content, err := os.ReadFile(pluginPath)
+		if err != nil {
+			return fmt.Errorf("clevis plugin %s: %v", pluginPath, err)
```
pilotstew: This is fatal for a build nobody opted into: `config.go` turns clevis on whenever the build is universal and `enable_clevis` is unset, so any universal build without the plugin beside the init binary dies here. Only the packaging paths this PR updates place it.

Only fail hard when someone actually asked for clevis:

```go
if conf.explicitEnableClevis {
    return fmt.Errorf("clevis plugin %s: %v", pluginPath, err)
}
warning("clevis plugin %s unavailable, building without clevis: %v", pluginPath, err)
conf.enableClevis = false
```

The `enableClevis = false` matters, otherwise `initConfig.EnableClevis` still says clevis and init goes looking for a plugin that isn't in the image.

Worth noting the fido2 plugin next door is best effort and defaults off, so a missing one there is harmless. Clevis defaulting on is what makes this fatal.

mikispag: Ugh - the earlier tests were clearly not adequate to catch this edge case. Thank you for catching it!

I've updated the implementation so we only fail hard when `enable_clevis: true` was explicitly configured in `booster.yaml` (`conf.explicitEnableClevis`). If Clevis was enabled implicitly (e.g. `universal: true` or auto-detection) and `clevisplugin.so` is missing, the generator now logs a warning, resets `conf.enableClevis = false`, and allows the build to succeed. I added proper unit tests covering both the implicit fallback and explicit failure paths.

By the way, I also removed the hardcoded `-ldflags="-s -w"` from `PKGBUILD` and `booster.spec` per your feedback so distro-standard debuginfo and stripping mechanisms work as intended.

pilotstew: I think the commit didn't make it into the push. This is still the unconditional `return`, with no `explicitEnableClevis` check or `warning()` in the block, and the `-ldflags="-s -w"` are still in all three packaging files too. A clean build with only `init` in the directory and `universal: true` still produces no image, and the integration suite still fails 110 of 115.

init/main.go
```diff
@@ -997,6 +997,12 @@ func boost() error {
 		return err
 	}
 
+	if config.EnableClevis {
+		if err := loadClevisPlugin(); err != nil {
```
pilotstew: This eager check can produce an image that does not boot and doesn't warn.

Under `CGO_ENABLED=0` the `!cgo` stub returns "clevis plugin requires CGO" unconditionally, whatever the image actually contains, and `boost()` returning here is terminal. The generator sets `enableclevis` from host detection or `universal: true` and cannot see how `init` was compiled, so the two disagree and the build stays clean. Booster's README builds init with `CGO_ENABLED=0`.

I booted two images differing only in how init was built, same config, both carrying `clevisplugin.so` and `enableclevis: true`:

```
CGO_ENABLED=1  ->  Hello, booster!
CGO_ENABLED=0  ->  Press ENTER to reboot        (nothing else printed)
```

The silence is the placement: this runs before `/dev` is mounted and `devKmsg` is opened, so `severe()` has nowhere to write.

Two changes fix it. Move the load below the `/dev` mount so a failure can be reported:

```go
	devKmsg, err = os.OpenFile("/dev/kmsg", unix.O_WRONLY, 0o600)
	if err != nil {
		return err
	}

	// After /dev/kmsg, so a failure here can be reported rather than dying mute.
	if config.EnableClevis {
		if err := loadClevisPlugin(); err != nil {
			return err
		}
	}
```

and in `clevis_nocgo.go`, distinguish "cannot load plugins at all" from "plugin is missing":

```go
func loadClevisPlugin() error {
	warning("clevis: init was built without cgo, clevis tokens cannot be unlocked")
	return nil
}
```

With both, the same static image boots:

```
booster: clevis: init was built without cgo, clevis tokens cannot be unlocked
Hello, booster!
```

The cgo path keeps `return err`, so a genuinely broken image still fails fast and now prints why. I verified the no-clevis-device case.

mikispag:
Ugh, git is hard. The working tree modifications were not staged into the git index when I ran `git reset --soft`, causing the previous index to be recommitted without the working tree edits. PTAL.

init/clevis_nocgo.go
```diff
@@ -0,0 +1,16 @@
+//go:build !cgo
+
+package main
+
+import (
+	"errors"
```
pilotstew: `errors` is unused now that the stub returns nil, so `CGO_ENABLED=0 go build ./init/` fails with "errors imported and not used". That is on me, I showed the function body and not the import that goes with it. Drop this line and both configurations build.

The test harness builds init with cgo, so the `!cgo` path is never built by the suite, which is the same configuration the boot fix exists for. Sorry I missed it.

pilotstew:
With the import fix this branch passes stress testing with the full test suite

init/clevis_nocgo.go
mikispag: Done.

pilotstew:
Looks good to me

PR #417, opened 2026-08-19
init/luks: pair a crypttab entry with a command line parameter that names the same device

pilotstew:
A crypttab entry and a kernel parameter can describe the same disk in different words. The entry says `LABEL=crypt`, the command line says `rd.luks.name=$UUID=root`. Booster treats those as two different devices today, because it compares how a device was written before what it resolves to.

Only one of the two then does anything. The command line is parsed before crypttab is read, so its mapping is the one that matches, and the entry's key file, its `tries=`, its `header=` are dropped without a message. The entry's mapping stays in the list matching nothing.

This is a follow-on to the per-device option work that just landed: that change made the option sources compose in one place, and this makes them compose for a device the two sources spell differently.

## What changes

Pairing now happens when the device appears, which is the first moment it can. Whether `LABEL=crypt` and a given UUID are the same disk is not knowable while parsing, which is why matching identical spellings was as far as the old code could go.

When more than one mapping answers for a device, they are folded into one. The command line keeps the volume name and the device reference. The crypttab entry contributes its options field and the key file it named. The result then goes through the same composition every other source goes through, so precedence is unchanged: `rd.luks.key=` still outranks the entry's key file, and a per-device `rd.luks.options=` still replaces the entry's options rather than adding to them. The absorbed mapping is removed, so nothing attempts to unlock the device twice.

## What it means

If your crypttab names a device by label, path or WWID while your bootloader names it by UUID, the crypttab entry's settings now apply instead of being silently ignored. Setups where both sources already used the same spelling behave exactly as before.

## Tests

Unit tests cover an entry naming the device by label, by path and by WWID, and a second pinning that command line precedence survives the fold.

### DIFF
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -262,9 +262,14 @@ the file by either being ran as root, or by being passed
 > this warning is irrelevant with x-initrd.attach
 
 
-Crypttab entries and `rd.luks.*` parameters with the same LUKS UUID are
+Crypttab entries and `rd.luks.*` parameters describing the same device are
 merged, with cmdline parameters taking precedence and unspecified options
-being added from crypttab.
+being added from crypttab. The two sources need not name the device the
+same way: an entry written `LABEL=`, `WWID=` or as a path is matched to a
+`rd.luks.*` parameter written `$UUID` once the device appears and both
+spellings are known to resolve to it. The volume name and device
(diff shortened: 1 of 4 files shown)


pilotstew:
Doing some more testing on this I'm not quite happy with the implementation. I'm going to close this and rework it.

Issue #70, opened 2021-03-27
Consider strip .smp_locks from ko


oech3:
`.smp_locks` and `.rela.smp_locks` seems related to Big Kernel Lock which is no longer exists Linux 2.6.37.
But I can't find evidence that they are really not used.

Size of btrfs.ko is reduced about 27KB by stripping them.

In my PC, modules loaded after boot are stripped (by my custom hook).
They don't have `.note.* .comment *bug_table .BTF *orc_unwind* *mcount_loc *smp_locks *bpf_raw_tp_map *ftrace*` sections.
And I don't see trouble.

anatol:
It worth trying to see how it works. Feel free to send a patch for it. I will merge it after 0.4 release.

oech3:
`*smp_locks` are exists in modules built with `CONFIG_UNUSED_SYMBOLS=n CONFIG_TRIM_UNUSED_KSYMS=y`

oech3:
I'm really sorry. It is related to SMP alternatives.
It is needed for non-SMP-capable system and CPU hotplug.

It may strippable for SMP-capable system optionally.

oech3:
Option like `strip_section: ,*smp_locks,*mcount_loc,*bpf_raw_tp_map,*ftrace*` would be nice.

Issue #36, opened 2021-02-20
strip kernel modules


oech3:
initramfs size is reduced by stripping kernel modules.
mkinitcpio has strip hook.

anatol:
Sounds like a good idea to decease size of the image.

And I think `strip` should be called for regular binaries as well.

anatol:
I am looking at the `strip` [hook from mkinitcpio](https://github.com/archlinux/mkinitcpio/blob/master/install/strip) and it looks like stripping is *not* run for kernel modules. It is run for regular binaries and shared/static libraries only.

```shell
    find "$BUILDROOT" -type f -perm -u+w -print0 2>/dev/null | while read -d '' bin; do
        case $(file -bi "$bin") in
            *application/x-sharedlib*)
                # Libraries (.so)
                strip --strip-unneeded "$bin"
                ;;
            *application/x-archive*)
                # Libraries (.a)
                strip --strip-debug "$bin"
                ;;
            *application/x-executable*)
                # Binaries
                strip --strip-all "$bin"
                ;;
        esac
    done
```

anatol:
So for kernel modules and shared libraries we need to use `--strip-unneeded` and for ELF binaries `--strip-all`. I have a change locally that seems work fine.

Stripping ELF binaries decreases size of the final image by about 10%. It is because Arch binaries already exclude debug symbols and there is not that much to strip.

anatol:
The functionality is implemented at `master` branch. PTAL and let me know if you see any issues.

PR #418, opened 2026-08-19
strip signed modules

pilotstew:
Issue #401 and #410 led me here. Referenced issues  #36, #57, #65, #69, #70.

`strip: true` silently destroys the signature on every signed kernel module it packs.

A module's signature is appended after the end of its ELF image. `strip` rewrites the ELF and discards everything past it, and exits successfully while doing so. The image looks fine and boots fine until the kernel enforces module signatures, at which point `finit_module` returns EKEYREJECTED and the boot stops:

```
booster: finit(virtio_scsi): key was rejected by service
booster: finit(ata_piix): key was rejected by service
```

Signing and enforcing are separate things, which is why this went unnoticed. `CONFIG_MODULE_SIG_ALL=y` signs every module and still loads unsigned ones happily. Refusal needs `CONFIG_MODULE_SIG_FORCE`, `module.sig_enforce=1`, or lockdown in integrity mode, which some distributions turn on under Secure Boot. So nothing happens on an Arch-family machine, and the same image will not boot on a Fedora one.

## What changes

Signed modules are no longer stripped. dracut's module loop already skips them, keyed on the same 28-byte marker.

Modules give up debug info only. Booster removed a fixed section list from every ELF it packed, and three of those are read by the kernel out of the module file rather than being debug data: the ORC unwind tables behind kernel backtraces, the BTF that registers the module under `/sys/kernel/btf`, and the build-id note. dracut strips modules with `-g` and nothing more, reserving the aggressive form for an explicit flag. This matches that.

A build-time warning when a module that was never signed, from DKMS or a local build, is packed for a kernel that will refuse it. Booster cannot fix that for the user, but it should not ship an image that dies at boot with nothing in the build output explaining why.

`strip_extra_sections` puts the aggressive behaviour back per section, for anyone who wants those bytes.

Userspace files are untouched by all of this, which is where stripping earns its keep: the init binary alone goes from 23.7 MB to 16.5 MB. The change costs roughly 200 KiB of module sections but makes the option safe to enable on machines that enforce signatures, which could not use it before.

## Testing

A unit test asserts a signed module reaches the image byte for byte, and that an unsigned one is still stripped, so the skip is not a blanket opt-out.

An integration test boots with `strip` enabled and `module.sig_enforce=1`, which makes the kernel refuse any module whose signature went missing, the same thing a Secure Boot machine does unasked. With the fix reverted the VM never reaches userspace and the console fills with the error above.

### DIFF
--- a/docs/manpage.md
+++ b/docs/manpage.md
@@ -70,6 +70,8 @@ serialize_tokens:
 
  * `strip` is a boolean flag that enables stripping of ELF files before adding them to the image. Binaries, shared libraries and kernel modules are examples of ELF files that get processed with the strip UNIX tool. The default is `false`. Signed kernel modules are skipped.
 
+ * `strip_extra_sections` is a comma-separated list of ELF sections to remove from kernel modules in addition to debug info, for example `*orc_unwind*, .BTF`. Names are passed to `strip -R`. Only applies when `strip` is set.
+
  * `extra_files` is a comma-separated list of extra files to add to the image. If an item starts with slash ("/") then it is considered an absolute path. Otherwise it is a path relative to /usr/bin. If the item is a directory then its content is added recursively. There are a few special cases:
     * adding `busybox` to the image enables an emergency shell in case of a panic during the boot process.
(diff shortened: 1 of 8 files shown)


pilotstew:
I meant to note somewhere that this PR brings booster in line with dracut's default handling of strip. I *believe* this *could* make strip: true a safe default. While `strip_extra_sections` becomes the "here be dragons" option

anatol:
LGTM, thank you!

PR #420, opened 2026-08-22
tests: repairs to the integration suite found by stress testing

pilotstew:
Nineteen repairs to the integration suite, all found by repeated full suite stress testing. The goal across this and two follow-up PRs is a suite that bootstraps itself, runs reliably, and runs fast.

This is meant to be merged after "init: five fixes found while stress testing the integration tests". Three tests here assert a log line that PR adds: TestCtxAwareFido2CancelMultiToken, TestCtxAwareFido2CancelOnFallback and TestSSHRemoteUnlockFido2Pending. Nothing else depends on it, and this branch is rebased off those commits so the diff here is tests only.

**Tests asserting things the code never promised.** Modules load concurrently, so two console lines can arrive in either order while ConsoleExpect only moves forward. The SCSI host number in a HWPATH depends on which storage driver registers first. A device name like /dev/sda depends on probe order once a test attaches a second disk, which is what made the hibernate test resume from a swap partition.

**Generators that failed quietly, or took root they did not need.** The ISO generator's sudo prompt fired the ERR trap, which deleted the staging tree while the rest of the script kept writing into it, so the real failure surfaced as a pile of missing-file errors. A sudo PAM prompt on a machine that authenticates with a hardware token was captured into a losetup command substitution, leaving a loop device attached to the finished image. The ESP images are now built with no root at all.

**Assets that depended on the host.** The large-image lz4 test named nvidia's modules, so the image never built on a machine without those packages. The zfs assets copied the host's /etc/zfs/zpool.cache, which holds nothing unless the host keeps pools there, and the guest then hung with no pool to import.

**Host resources assumed to be free.** The ssh forward was hardcoded to port 10022, so any two tests using it could never run at the same time.

The last two commits bump vmtest and pass its new QemuOptions.Output field a per-test writer, so each guest's console goes through t.Log and a failing VM's output is attributed to the test that produced it.


anatol:
Woah, thank you! I will look through the changes.

pilotstew:
I have quite a lot staged and some of it conflicts. So I'm pushing things that don't conflict and in this case making the test suite easier and faster will hopefully help with reviews, especially since there seems to be more contributions lately. Hopefully you're not getting overwhelmed. Take your time. @anatol

anatol:
Thank you once again.

Integration tests is something that is very important for this project. And at the same time these tests has always had some sort of corner-cases and quirks. Fixing it would will improve testability of the project and make the development much more enjoyable.

Any improvements to testing infrastructure are very welcome.

PR #423, opened 2026-09-02
generator: resolve a library by architecture, not by name alone

pilotstew:
See #421 

DT_NEEDED entries are resolved against a fixed directory list and the first filename hit wins, with no check that what was found is loadable by the object that asked for it. A soname does not identify a file on a multilib host, where the same name exists once per architecture.

On Fedora /usr/lib holds the 32-bit libraries and is searched ahead of /usr/lib64, so an image built there for a 64-bit init receives a 32-bit libc and no 64-bit one at all. The dynamic loader then refuses to start init and exits 127, and because that process is PID 1 the kernel panics with "Attempted to kill init! exitcode=0x00007f00". Fedora is the distro exposed to this because its spec builds init with cgo for the FIDO2 plugin, so init is dynamically linked there and the image has to carry a working libc.

elfPath now takes the requesting object's ELF class and machine and skips candidates that do not match. Reordering the search list would fix x86-64 Fedora only: it inverts on a 32-bit host and does nothing for the per-architecture directories used by multiarch layouts. A library with no matching candidate now fails the build rather than producing an image that cannot boot.

Verified two ways. On a Fedora 44 container with glibc.i686 installed the dependency walk packed the 32-bit libc, the 32-bit loader and no 64-bit libc before the change, and the correct 64-bit set after, with the image's own loader resolving init cleanly. Under QEMU with the 32-bit directory forced ahead of the 64-bit one, the boot panicked with that exitcode before the change, and after it unlocks LUKS, mounts the root filesystem and reaches switch_root. A user who hit this on Fedora 44 rebuilt with the change and confirmed the image boots.

### DIFF
--- a/generator/image.go
+++ b/generator/image.go
@@ -325,7 +325,9 @@ func elfSectionContent(s *elf.Section) (string, error) {
 var elfLibDir = []string{"/usr/lib", "/lib", "/usr/lib64", "/usr/lib/x86_64-linux-gnu"}
 
 // for a given library (e.g. libc.so.6) finds absolute path to the library file
-func elfPath(lib string) string {
+// loadable by an object of this class and machine. A multilib host carries the
+// same soname once per architecture, so the name alone does not identify a file.
+func elfPath(lib string, class elf.Class, machine elf.Machine) string {
 	if filepath.IsAbs(lib) {
 		return lib
 	}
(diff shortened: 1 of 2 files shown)


anatol:
LGTM!

PR #425, opened 2026-09-13
Testing bootstrap

pilotstew:
Currently the test suite builds each disk image the first time a test needs it. The goal here was to create a simpler path for new contributors to build the assets list tools required via a single sudo once before starting the test suite.

`go test ./tests -bootstrap -timeout=0` builds every missing image behind a single prompt and exits without running tests. `timeout=0` is is recommended for the large number of assets to be built/downloaded.

Each generator declares its own required tools in a `# tools:` line. An image whose tools are absent is named and skipped instead of failed, and the run ends with counts of built, present, skipped, failed and broken plus a tick list of the tools that would unlock the rest.

The second commit is unrelated to bootstrapping but surfaced under it: `btrfs_raid0.sh` was the one generator that attached a loop device and never detached it, so a run that builds every image left an orphan bound to a deleted file.


anatol:
looks good!

PR #428, opened 2026-09-20
fix: preserve env when execing init

cootshk:
Envrionment variables aren't properly passed into init systems when using booster.

You can see this by running `tr '\0' '\n' < /proc/1/environ` as root.

This causes OpenRC to not display color, as it is looking for the TERM=linux variable.

Tested on my Artix Linux machine.


anatol:
Cc @pilotstew

pilotstew:
I wouldn't use os.Environ() directly here. It reads the live environment and picks up booster's own PATH=/usr/bin. openrc-init and busybox init overwrite PATH themselves but the value is the initramfs's lookup path, missing /usr/sbin. 

dracut and systemd both shipped the live-environment form historically and both changed it for the same reason. dracut was passing its own PATH along while unsetting HOME and TERM, later replaced that with the kernel's values, quoting envp_init out of init/main.c. systemd used a plain execv for a foreign init then switched it to execve with saved_env, a snapshot taken at startup, once PID 1 had begun unsetting HOME for its own purposes. mkinitcpio is more restrictive.

Snapshot/saved_env at start concept for booster would be kernelEnv = os.Environ() as main()'s first statement, ahead of the Setenv, then exec kernelEnv. This gives just the base kernel environ without any extra taint. 

testing: TestAlpineLinux didn't catch this, since Alpine's /sbin/init is busybox and it forces TERM=vt102 on a serial console whether TERM is unset or linux. The same image with TERM=vt100 on the cmdline does reproduce it, no escape sequences before the change and colored OpenRC output after.

